PR #13, opened 2023-12-01
fix(back): Fix getting category name in COCO Importer

magwyz:
The category_name field does not always exist in an annotation object.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -37,6 +37,7 @@ All notable changes to Pixano will be documented in this file.
 - Fix pip commands in notebooks for Google Colab (pixano#11)
 - Fix broken link in CHANGELOG (pixano#4)
 - Fix documentation website API reference generation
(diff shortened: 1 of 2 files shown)


magwyz:
I got this issue from a dataset annotated with CVAT.

cpvannier:
Thank you for the suggestion!

I left a few notes on what I would change, just a minor edit for readability.

Also, it would be great if you could format your code with Black so that it passes our lint test. You can install the [VS Code extension](https://marketplace.visualstudio.com/items?itemName=ms-python.black-formatter) and set it to format files automatically on save, or if you use another editor, you can run it directly with the [PyPI package](https://pypi.org/project/black/).

And finally, this update should be mentioned in the CHANGELOG.md file under `Unreleased` - `Changed`:
```
- Update COCO Importer to get the category name from the "categories" field as it does not always exist in the "annotations" field
```

Thanks again!

BertrandRenault (approved):

jrabary (approved):

PR #30, opened 2024-01-10
feat(front): Select models if more than one and display errors

timothee-LJN:
## Description

Fixes pixano/pixano-project-manager#7

### DIFF
--- a/ui/apps/pixano/src/routes/+layout.svelte
+++ b/ui/apps/pixano/src/routes/+layout.svelte
@@ -17,8 +17,6 @@
   let currentDatasetName: string;
 
   async function handleGetModels() {
-    console.log("App.handleGetModels");
-
     models = await api.getModels();
     modelsStore.set(models);
   }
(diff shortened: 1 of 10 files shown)


cpvannier (changes requested):
When selecting one of the smart tools, I see in the console that the app is trying to load the embeddings of the first model in the list before I make a selection.

And when I select the right model, I get a "No embeddings found" modal but I can see in the console that its embeddings have been loaded.

cpvannier (approved):
Minor details:
- After selecting a model within a dataset, when opening another dataset, the app will prompt the user to select a model directly instead of waiting for them to click on a smart tool first.
- When selecting a model without embeddings, `api.getItemEmbeddings` returns a 404 but `modelApi.loadEmbeddings` still tries iterating on `item.embeddings` and encounters a TypeError as it is empty. Maybe a `if (item.embeddings)` should be added instead of just checking `if (item)`.

PR #33, opened 2024-01-15
feat: Load dynamic image parameters

timothee-LJN:
## Description

Fixes pixano#25


cpvannier (approved):

cpvannier:
Note:
Can create new attributes on scene level from suggested ones (number of people, is group...), but cannot do that on object level attributes (422 Unprocessable Entity)

cpvannier (changes requested):
Can create new attributes on scene level from suggested ones (number of people, is group...), but cannot do that on object level attributes

Getting "422 Unprocessable Entity" in the server logs with no Python errors

cpvannier (approved):

BertrandRenault (approved):

PR #41, opened 2024-01-18
fix(front): Improve multi-view and tool selection

timothee-LJN:
## Description

Several small bug fixes / improvements:
- The point radius and polygon lines are now correlated with the zoom factor
- Bug fix on multiview for display and mask creation.
- Color change to harmonize creation tools (both rectangle and polygons).
- Modal to allow saving when switching items without saving.
- Display of a return button when the item does not exist.
- Display of "confidence" in tooltips only for model runs (not for ground truth).

Fixes pixano/pixano-project-manager#26
FIxes pixano/pixano-project-manager#21

### DIFF
--- a/ui/apps/pixano/src/components/layout/DatasetHeader.svelte
+++ b/ui/apps/pixano/src/components/layout/DatasetHeader.svelte
@@ -22,10 +22,14 @@
 
   import pixanoLogo from "@pixano/core/src/assets/pixano.png";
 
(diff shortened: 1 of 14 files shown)


cpvannier (commented):
I am not seeing tooltips for confidence values on model run objects, not sure why.
All the rest is working fine for me.

cpvannier (approved):

BertrandRenault (approved):

PR #53, opened 2024-01-22
fix(front): Load embeddings only once a model is selected

timothee-LJN:
fixes : pixano/pixano-project-manager#30

improve number of time we hit the api for : 
- loading new item
- loading embeddings


BertrandRenault (approved):

cpvannier (commented):
With the dropdown list, If I click select one model, then select another, the LoadModelModal disappears before I can confirm my selection.

timothee-LJN:
@cpvannier it's corrected thanks for your feedback

cpvannier (approved):

PR #55, opened 2024-01-22
fix(back): use uuid instead of imported id, for items and objects. keep original id

### LINKED ISSUE #50: Error when using COCOImporter with multiple splits.
I am using a COCO format dataset, created using supervision library:

image

When passing the splits as follows, the library does not behave as expected: `splits = ["test", "train", "valid"]`.

We notice the following issues in the explorer page:
- The id values for all images are set to 0: 
image
- The preview shows one image, but the one shown in full view is different.
- The mask data is used from the thumbnail mentioned above and not the displayed image.
- The splits shown in overview page does not correspond to the image shown in full view.

We have validated the images in another tool and noticed no such behaviour, and we get expected results when only using one split at a time.

Kindly investigate any potential issues with loading multiple splits, perhaps with how the id values are handled.

BertrandRenault:
This is a draft for using shortuuid instead of original id in coco_importer, to ensure id unicity

- [x] keep original id
- [x] ensure objects id unicity too
- [x] apply to all importers
- [x] fix tests

should fix #50

### DIFF
--- a/pixano/data/importers/coco_importer.py
+++ b/pixano/data/importers/coco_importer.py
@@ -55,11 +55,15 @@ class COCOImporter(Importer):
         tables = super().create_tables(
             media_fields,
             object_fields={
+                "original_id": "str",
                 "bbox": "bbox",
                 "mask": "compressedrle",
                 "category": "str",
             },
         )
+
+        # Add original id in main table
+        tables["main"][0].fields["original_id"] = "str"
 
         # Create categories
         categories = []
@@ -92,10 +96,6 @@ class COCOImporter(Importer):
             ) as f:
                 coco_instances = json.load(f)
 
-            # ensure uniqueness by using shortid
-            for im in coco_instances["images"]:
-                im["uuid"] = shortuuid.uuid()
-
             # Group annotations by image ID
             annotations = defaultdict(list)
             for ann in coco_instances["annotations"]:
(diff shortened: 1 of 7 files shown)


BertrandRenault:
Some tests are based on id (original data id), so generating a shortuuid id breaks theses tests.
Maybe we can add a "field" parameter in dataset.load_item() to select by 'original_id' instead of 'id' ?

cpvannier:
It is probably best to access the table directly with Lance to retrieve the generated UUIDs.

This will allow calling the endpoints in the same way that the frontend does, which is what we want to validate with these tests.

timothee-LJN (approved):

pixano/data/importers/coco_importer.py
```diff
@@ -53,12 +55,16 @@ def __init__(
...
             object_fields={
+                "original_id": "str",
                 "bbox": "bbox",
                 "mask": "compressedrle",
                 "category": "str",
             },
         )
 
+        # Add original id in main table
+        tables["main"][0].fields["original_id"] = "str"
```
cpvannier, line 66: Maybe add "original_id" to base importer (importer.py, line 104), instead of adding it in each importers (COCO, DOTA, Image) manually.

cpvannier (approved):

PR #58, opened 2024-01-25
feat(front): Create pre-annotation

timothee-LJN:
This PR integrates the Pre-annotation feature: 
- User can switch the Pre-annotation space feature on and off
- User can filter objects in the Pre-annotation table with a confidence slider
- User can reject or accept objects from the Pre-annotation table
- Current object selected in the Pre-annotation space is highlighted on ImageCanvas

How the data is processed:
- The Pre-annotation table is selected with source="Pre-annotation" in db.json (ItemObject.source_id)
- Accepted objects are flagged with review_state="accepted" in Pre-annotation table and copied to Ground Truth table
- Rejected objects are flagged with review_state="rejected" in Pre-annotation table
- Objects flagged as accepted or rejected are no longer suggested in the Pre-annotation space

image

### DIFF
--- a/ui/components/canvas2d/src/Canvas2D.svelte
+++ b/ui/components/canvas2d/src/Canvas2D.svelte
@@ -357,6 +357,9 @@
             //update visibility & opacity
             bboxKonva.visible(bboxes[i].visible);
             bboxKonva.opacity(bboxes[i].opacity);
+            const rect: Konva.Rect = bboxKonva.findOne(`#rect${bboxes[i].id}`);
+            const currentStrokeWidth = rect.strokeWidth();
+            rect.strokeWidth(bboxes[i].strokeFactor * currentStrokeWidth);
             //update color
             const style = new Option().style;
             style.color = colorScale(bboxes[i].id);
--- a/ui/components/canvas2d/src/Canvas2D.svelte
+++ b/ui/components/canvas2d/src/Canvas2D.svelte
@@ -1073,7 +1073,7 @@
 </script>
 
 <div
-  class={cn("flex h-full w-full bg-slate-100 transition-opacity duration-300 delay-100", {
+  class={cn("flex h-full w-full bg-gray-900 transition-opacity duration-300 delay-100", {
     "opacity-0": !isReady,
   })}
   bind:this={stageContainer}
--- a/ui/components/canvas2d/src/Canvas2D.svelte
+++ b/ui/components/canvas2d/src/Canvas2D.svelte
@@ -1081,7 +1081,7 @@
 </script>
 
 <div
-  class={cn("flex h-full w-full bg-gray-900 transition-opacity duration-300 delay-100", {
+  class={cn("flex h-full w-full bg-slate-800 transition-opacity duration-300 delay-100", {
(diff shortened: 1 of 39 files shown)


ui/components/canvas2d/src/Canvas2D.svelte
```diff
@@ -1073,7 +1075,7 @@
 </script>
 
 <div
-  class={cn("flex h-full w-full bg-slate-100 transition-opacity duration-300 delay-100", {
+  class={cn("flex h-full w-full bg-gray-900 transition-opacity duration-300 delay-100", {
```
cpvannier, line 1078: `Canvas2D.svelte:1078`, `ImageCanvas.svelte:59`: `bg-slate-800` seems closer to the color used in the Sketch template than `bg-gray-900`, and would be more coherent with the rest of the app's color palette (92 uses of slate vs 7 uses of gray).

The remaining uses of gray could also be updated to use slate instead.

cpvannier:
Regarding the items that require more specifications:
- **Objects from the first model run are added to the pre-annotation pile:** This should be changed to only add objects with `source_id="Pre-annotation"`. To test the feature, the `source_id` of an objects table can be easily changed by changing the value of `tables.objects.[table_name].source` in the DatasetInfo file, db.json.
- **Accepted annotations are changed to "Ground truth":** I think this is the correct way.
- **Rejected annotation are deleted:** I think it's either that or tagging it with a feature like `rejected=true`. I would probably go for deleting as I don't see the point in keeping the annotation if it is rejected, but the choice is not up to me.

timothee-LJN:
@cpvannier after a talk with @jrabary `pre-annotation` objects will be copied in ground truth AND flagged `accepted` or `rejected` so we can track job done and remaining!  I'll add these flag in this PR.

timothee-LJN:
The PR seems ready in the frontend but I cannot saved. Got this error in the backend. It only occurs in the dataset that I changed to test this feature.  
image

cpvannier:
> The PR seems ready in the frontend but I cannot saved. Got this error in the backend. It only occurs in the dataset that I changed to test this feature. image

Should be fixed by 1be3c9291a2ae2aea6df54451e4bdfa936512533 which changes the default null value for numeric features. This is a problem with our lance dependency that is tracked in pixano/pixano-project-manager#27.

ui/components/imageWorkspace/src/components/SaveShape/SaveShapeForm.svelte
```diff
@@ -96,60 +79,12 @@
...
-  };
-  $: {
-    const result = objectValidationSchema.safeParse(objectProperties);
-    isFormValid = result.success;
-  }
 </script>
 
 {#if shape.status === "inProgress"}
   <form class="flex flex-col gap-4 p-4" on:submit|preventDefault={handleFormSubmit}>
     <p>Sauvegarde {shape.type}</p>
```
BertrandRenault, line 86: Sauvegarde -> Save

BertrandRenault (approved):

cpvannier (approved):

PR #87, opened 2024-02-02
feat(front): Create fixed size sections for Ground truth and Model run 

### LINKED ISSUE #80: Improve object inspector look and feel
- Ground Truth and model run panel are the same size
- GT and MR don't close
- One can resize panels size

image

timothee-LJN:
fixes #80 

**Change description**
- Ground Truth and model run panel are the same size
- GT and MR don't close
- Model are selected from a dropdown.
- object are closed by default
- refactoring to get a more consistant naming

**question** 
@jrabary Model run objects are filtered only in the inspector, which means that they are still visible on the image and toggling visibility is for all model run. Shall I change that? (it would make sense I guess, but I prefer to get confirmation before coding). 

**not done**
- Sections size are draggable. @jrabary this is a quite complex task which would need to add an external library (I'd reckon). Do you confirm that it's needed? 

image


timothee-LJN:
**update :** 

following a call with lead design Janro, the following changes have been added : 
- focus on pre annotation when its active: GT is hidden and the form takes the all remaining space
- object card icons are hidden expect when user is hovering the card or when the card is open
- fontWeights have been harmonized

cpvannier (commented):
With the new drop-down selection for "Model run", maybe only the selected model should be displayed on the image?
The visibility toggle next to "Model run" should probably only applies to the selected model as well. 

image

cpvannier:
Also with a small window, spacing between Ground truth and Model run is correct:
image


But with a taller window, empty space appears between the two that should be used for displaying more Ground truth items: 
image

BertrandRenault (approved):

timothee-LJN:
@cpvannier I've made the changes your requested.

cpvannier (approved):

PR #91, opened 2024-02-05
feat(back): Choose Pre-annotation or Model run in InferenceModel

### LINKED ISSUE #90: Add option to save annotations as "Pre-annotation" or "Model Run"

cpvannier:
## Description

Fixes #90 

### Added

- Add option in pre-annotation notebook to save annotations as "Pre-annotation" or "Model run"
  - The `process_type` argument in `InferenceModel.process_dataset()` is now split from `"obj"` to either `"pre_ann"` or `"model_run"` to let the user choose if they want the generated annotations to appear as "Pre-annotation" or "Model run" in the app.
  - The pre-annotation notebook is also updated to reflect this new choice.

### Fixed

- Update "Ignore" pre-annotation button to "Reject" in the app
  - As discussed during the latest sprint demo (Fri 02/02/2024), "Reject" makes more sense for the end users, and better reflects the fact that the annotation will not be shown again.
  - This update also improves consistency with new documentation in inference_model.py and pre_annotation.ipynb describing the choice between "Pre-annotation" and "Model run".
- Update Prettier config for 3.2.5
  - Prettier 3.2.5 makes a change to the parser used for tsconfig.json files ([more info in their changelog](https://github.com/prettier/prettier/blob/main/CHANGELOG.md#use-json-parser-for-tsconfigjson-by-default-16012-by-sosukesuzuki)). Update the Prettier config to keep formatting those files with the jsonc parser.

### DIFF
--- a/pixano/models/inference_model.py
+++ b/pixano/models/inference_model.py
@@ -179,9 +179,9 @@ class InferenceModel(ABC):
     def process_dataset(
         self,
         dataset_dir: Path,
-        process_type: str,
-        ann_type: str,
         views: list[str],
+        process_type: str,
+        ann_type: str = None,
         splits: list[str] = None,
         batch_size: int = 1,
         threshold: float = 0.0,
@@ -190,6 +190,7 @@ class InferenceModel(ABC):
 
         Args:
             dataset_dir (Path): Dataset directory
+            views (list[str]): Dataset views
             process_type (str): Process type
                                 - 'ann' for annotations
                                 - 'segment_emb' for segmentation embeddings
@@ -197,7 +198,6 @@ class InferenceModel(ABC):
             ann_type (str): Annotation type
                                 - 'Pre-annotation' for pre-annotations to accept or reject as Ground Truth
                                 - 'Model run' for annotations to compare to Ground Truth
-            views (list[str]): Dataset views
             splits (list[str], optional): Dataset splits, all if None. Defaults to None.
             batch_size (int, optional): Rows per process batch. Defaults to 1.
             threshold (float, optional): Confidence threshold for predictions. Defaults to 0.0.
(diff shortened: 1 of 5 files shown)


pixano/models/inference_model.py
```diff
@@ -62,16 +62,20 @@ def __init__(
...
             process_type (str): Process type
-                                - 'obj' for preannotation
-                                - 'segment_emb' for segmentation embedding precomputing
-                                - 'search_emb' for semantic search embedding precomputing
+                                - 'ann' for annotations
+                                - 'segment_emb' for segmentation embeddings
+                                - 'search_emb' for semantic search embeddings
+            ann_type (str): Annotation type
+                                - 'Pre-annotation' for pre-annotations to accept or reject as Ground Truth
+                                - 'Model run' for annotations to compare to Ground Truth
```
BertrandRenault, line 78: Don't you think it would be easier to have 4 process types instead of 3 process_types and 2 annotation_type ?
process_type (str): Process type
  - 'pre_ann' for preannotations
  - 'model_run' for model run annotations
  - 'segment_emb' for segmentation embeddings
  - 'search_emb' for semantic search embeddings

cpvannier: Yes, I should have seen that. Merged `ann_type` with `process_type` in 48eba9c44f6681ca5295da59142572dda6eebf0a.

BertrandRenault (approved):

ui/.prettierrc.cjs
```diff
@@ -4,4 +4,12 @@ module.exports = {
   trailingComma: "all",
   printWidth: 100,
   plugins: ["prettier-plugin-svelte"],
+  overrides: [
```
timothee-LJN, line 7: question: what are those lines for? thanks!

timothee-LJN (approved):

ui/.prettierrc.cjs
BertrandRenault: tsconfig.json files are not strict json, it's "json with comments" (see prettier changelog link in Fixed > Update Prettier in description above). They choose to restrict this, but with these settings we can continue to have comments in tsconfig files (there is some comments inside)

PR #92, opened 2024-02-05
fix(api): Create bounding box from mask automatically

### LINKED ISSUE #89: Create bbox from mask and display it after saving
- [x] Stop displaying an empty bounding box before it's created from a mask or polygon (#92)
- [ ] Display the bounding box created from the mask or polygon as soon as the item is saved (#131)

cpvannier:
## Description

Part 1/2 of issue #89 

### Fixed

- Fix creating bounding box from mask automatically
  - Create bounding box from mask when saving an object without needing to create an empty box in the frontend first. This fixes the issue where the empty box would be displayed on the image while it shouldn't.

### DIFF
--- a/pixano/data/item/item_object.py
+++ b/pixano/data/item/item_object.py
@@ -96,12 +96,8 @@ class ItemBBox(BaseModel):
             else None
         )
 
-    def to_pyarrow(self, mask: ItemURLE = None) -> BBox:
+    def to_pyarrow(self) -> BBox:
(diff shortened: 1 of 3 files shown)


cpvannier:
Waiting on #91 to be merged to fix the frontend formatting check (dbc4b8e2df81d62c65129beb7b9bab51c16dd4e8)

timothee-LJN (approved):

PR #93, opened 2024-02-05
fix: S3 and URL wrongly handled

### LINKED ISSUE #95: Fix S3 support
Regressions:
- an environment variable required by lancedb for S3 (AWS_ENDPOINT renamed ENDPOINT_URL)
- URL images doesn't work anymore

BertrandRenault:
Fixes #95

- AWS_ENDPOINT and AWS_REGION are required by lancedb for S3, AWS_ENDPOINT was previously renamed to ENDPOINT_URL.
- URL images were not correctly handled

### DIFF
--- a/ui/components/canvas2d/src/Canvas2D.svelte
+++ b/ui/components/canvas2d/src/Canvas2D.svelte
@@ -204,7 +204,11 @@
     for (const view of Object.values(selectedItem.views)) {
       zoomFactor[view.id] = 1;
       images[view.id] = new Image();
-      images[view.id].src = `/${view.uri}` || view.url;
+      if (utils.isValidURL(view.uri)) {
+        images[view.id].src = `${view.uri}` || view.url;
+      } else {
+        images[view.id].src = `/${view.uri}` || view.url;
+      }
       images[view.id].onload = () => {
         // Find existing Konva elements in case a previous item was already loaded
         if (currentId) {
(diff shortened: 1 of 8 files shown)


cpvannier (approved):

ui/components/core/src/utils.ts
```diff
@@ -21,3 +21,11 @@ import { schemeSet3 } from "d3-scale-chromatic";
 export function ordinalColorScale(range: Iterable<string>) {
   return scaleOrdinal(schemeSet3).domain(range);
 }
+
```
timothee-LJN, line 24: in order to better organize utils, there is a dedicated folder `core/src/lib/utils`. Maybe we could create an index file which would allow to do a `utils.fooFunction`?

ui/components/canvas2d/src/Canvas2D.svelte
```diff
@@ -204,7 +204,11 @@
     for (const view of Object.values(selectedItem.views)) {
       zoomFactor[view.id] = 1;
       images[view.id] = new Image();
-      images[view.id].src = `/${view.uri}` || view.url;
+      if (utils.isValidURL(view.uri)) {
+        images[view.id].src = `${view.uri}` || view.url;
```
timothee-LJN, line 208: I think `|| view.url` was added here during data migration to avoid bugs but should probably not be needed in the future?

timothee-LJN (approved):

ui/components/canvas2d/src/Canvas2D.svelte
BertrandRenault: agreed, but in a future PR

PR #108, opened 2024-02-07
feat(front): Sort preannoted objects by axis

timothee-LJN:
## issue
#99 

## description
Pre annotation objects are now sorted based on their X position (`bbox.coords[0`)

### DIFF
--- a/ui/components/imageWorkspace/src/lib/api/objectsApi.ts
+++ b/ui/components/imageWorkspace/src/lib/api/objectsApi.ts
@@ -156,9 +156,9 @@ export const sortAndFilterObjectsToAnnotate = (
       return confidence >= confidenceFilterValue[0];
     })
     .sort((a, b) => {
-      const confidenceA = a.bbox?.coords[0] || 0;
-      const confidenceB = b.bbox?.coords[0] || 0;
-      return confidenceA - confidenceB;
+      const firstBoxXPosition = a.bbox?.coords[0] || 0;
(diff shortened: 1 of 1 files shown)


cpvannier (approved):

ui/components/imageWorkspace/src/lib/api/objectsApi.ts
```diff
@@ -156,9 +156,9 @@ export const sortAndFilterObjectsToAnnotate = (
       return confidence >= confidenceFilterValue[0];
     })
     .sort((a, b) => {
-      const confidenceA = a.bbox?.confidence || 0;
-      const confidenceB = b.bbox?.confidence || 0;
-      return confidenceB - confidenceA;
+      const confidenceA = a.bbox?.coords[0] || 0;
+      const confidenceB = b.bbox?.coords[0] || 0;
+      return confidenceA - confidenceB;
     });
```
BertrandRenault, line 162: "confidenceA/B" var name is not relevant anymore. But maybe both sort should be available later ?

BertrandRenault (approved):

ui/components/imageWorkspace/src/lib/api/objectsApi.ts
timothee-LJN: I've rename the var for now. 
Allowing both sorts would have quite big impact on the UI so we decided to keep that feature for later. It would be difficult then to redo this sort function !

PR #117, opened 2024-02-08
fix(front): Fix buttons in ImageInspector

### LINKED ISSUE #116: Fix buttons in ImageInspector
- Some tooltips are missing (object visibility, object highlighting, object card toggle, bounding box and mask toggles)
- Existing tooltips are inconsistent (first letter should be uppercase)
- As of #87, object cards are closed by default, so editing button should now open object card as well

cpvannier:
## Issue

Fixes #116 

## Description

- Open object card when clicking on editing button (since as of #87, object cards are closed by default)
- Add missing button tooltips (scene features, object visibility, object highlighting, object card toggle, bounding box and mask toggles)
- Fix existing button tooltips consistency (first letter should be uppercase)

### DIFF
--- a/ui/components/imageWorkspace/src/components/ImageInspector/ObjectCard.svelte
+++ b/ui/components/imageWorkspace/src/components/ImageInspector/ObjectCard.svelte
@@ -129,6 +129,7 @@
       <button
         class="rounded-full border w-3 h-3 mr-2 flex-[0_0_0.75rem]"
         style="background:{color}"
+        title="Highlight object"
         on:click={onColoredDotClick}
       />
       <span class="truncate w-max flex-auto">{itemObject.id}</span>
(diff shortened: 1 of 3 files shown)


BertrandRenault (approved):

timothee-LJN (approved):
To ensure consistency I'd reckon the force the capitalization of the `tooltipIconContent` with the following 
```
    <Tooltip.Content>
      <p class="first-letter:uppercase">foo</p>
    </Tooltip.Content>
```
in `TooltipIconButton.svelte` (line 39)

cpvannier:
@timothee-LJN Adding `class="first-letter:uppercase"` to the <p> element inside <Tooltip.Content> seems to produce weird issues for padding.

Without `class="first-letter:uppercase"` :
image

With `class="first-letter:uppercase"` :
image

Quite strange...

timothee-LJN:
this weird indeed. I don't reproduce on my computer. I've no idea where it could come from!

cpvannier:
Ok, it's just a Firefox issue. It works as expected in Chrome.

The `<p>` element size if defined as if the whole word was in uppercase instead of just the first letter.

I am going to merge this PR as is for now then.

Issue #125, opened 2024-02-08
Ask confirmation before leaving item page


cpvannier:
Dataset navigation buttons (Previous item, Next item) ask for confirmation before changing the selected item.

But clicking on any other buttons (Pixano logo, Back button, Dashboard button, Dataset button) does not ask for confirmation first.

PR #126, opened 2024-02-08
feat(front): Ask for confirmation before leaving item page

cpvannier:
## Issue

#125 

## Description

When using any buttons in the DatasetHeader, ask for confirmation if they are any unsaved changes.
Previously, only the Next item and Previous item buttons would ask for confirmation.

### DIFF
--- a/ui/apps/pixano/src/components/layout/DatasetHeader.svelte
+++ b/ui/apps/pixano/src/components/layout/DatasetHeader.svelte
@@ -40,8 +40,8 @@
   let currentItemId: string;
   let isLoading: boolean;
   let canSaveCurrentItem: boolean;
-  type Direction = "previous" | "next" | "none";
-  let showConfirmModal: Direction = "none";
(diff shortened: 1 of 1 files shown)


ui/apps/pixano/src/components/layout/DatasetHeader.svelte
```diff
@@ -85,12 +81,24 @@
     return event.key;
   };
 
-  const handleConfirmClick = async () => {
+  const handleSaveandContinue = async () => {
```
timothee-LJN, line 84: there is a typo here: `handleSaveAndContinue`

timothee-LJN (approved):

ui/apps/pixano/src/components/layout/DatasetHeader.svelte
cpvannier: Thanks, fixed!

BertrandRenault (approved):

PR #127, opened 2024-02-08
feat: Add features auto-completion

### LINKED ISSUE #76: Add autocompletion for features edition (object/scene)
Autocomplete field while user is typing the input.

BertrandRenault:
close #76 

Notes: 
- for Scene features, as there is often/always a default value, the auto-completion only show the current option, others options are shown if we delete current value
image
image

- available values are read from tables in back, not from db.json. This could be done in a future PR, if necessary
- (back) merge with "categories" -> related to previous, if we choose to save available values in db.json, we will refactor "categories" to be more generic


ui/components/core/src/lib/types/datasetTypes.ts
```diff
@@ -139,6 +140,19 @@ export interface ItemFeature {
   required?: boolean;
 }
 
+export interface FeaturesAvailableValues {
+  [key: string]: {
+    [innerKey: string]: Array<string>;
```
timothee-LJN, line 145: your index here is not a variable, I think its more idiomatic to call always the same 

```

export interface FeaturesAvailableValues {
  [key: string]: {
    [key: string]: Array<string>;
  };
}
```

ui/components/imageWorkspace/src/ImageWorkspace.svelte
```diff
@@ -104,6 +109,30 @@
...
+          outputFeaturesAvailableValues[table] = {};
+          for (let feat of inputFeaturesAvailableValues[table]) {
+            if (feat.values.length) {
+              outputFeaturesAvailableValues[table][feat.name] = feat.values;
+            }
+          }
+        }
+        itemFeaturesAvailableValues.set(outputFeaturesAvailableValues);
+      })
+      .catch((err) => console.log(err));
```
timothee-LJN, line 134: maybe used `console.error(err)` here?

ui/components/imageWorkspace/src/ImageWorkspace.svelte
```diff
@@ -104,6 +109,30 @@
...
+    // but right now, we compute available feats only in getDataset (no S) route, used only for stats, and here
+    // later, we should get itemFeaturesAvailableValues directly from DatasetInfo["available_feat_values"]
+    api
+      .getDataset(currentDatasetId)
+      .then((datasetWithFeats) => {
+        const inputFeaturesAvailableValues: InputFeaturesAvailableValues =
+          datasetWithFeats.available_feat_values || {};
+        let outputFeaturesAvailableValues: FeaturesAvailableValues = {};
+        //convert from InputFeaturesAvailableValues to FeaturesAvailableValues
+        for (let table in inputFeaturesAvailableValues) {
```
timothee-LJN, line 124: Following our conversation, you could have used a reduce like so : 
```
        const outputFeaturesAvailableValues: FeaturesAvailableValues = Object.entries(
          inputFeaturesAvailableValues,
        ).reduce((tables, [tableName, tableValues]) => {
          tables[tableName] = tableValues.reduce(
            (features, feature) => {
              if (feature.values) {
                features[feature.name] = feature.values;
              }
              return features;
            },
            {} as FeaturesAvailableValues[""],
          );
          return tables;
        }, {} as FeaturesAvailableValues);
```

I personnaly tend to prefer reduce but a for in loop works also well!

ui/components/imageWorkspace/src/components/Features/FeatureFormInputs.svelte
```diff
@@ -99,13 +99,21 @@
         step={feature.type === "int" ? "1" : "any"}
         autofocus={i === 0 ? true : false}
         value={initialValues[feature.name]?.value || ""}
+        list="obj_availableValues_{feature.name}"
```
timothee-LJN, line 102: use lowerCamelCase !

ui/components/imageWorkspace/src/components/Features/FeatureFormInputs.svelte
```diff
@@ -99,13 +99,21 @@
...
         on:keyup={(e) => e.stopPropagation()}
         on:input={(e) =>
           handleInputChange(
             feature.type === "str" ? e.currentTarget.value : Number(e.currentTarget.value),
             feature.name,
           )}
       />
+      <datalist id="obj_availableValues_{feature.name}">
```
timothee-LJN, line 110: use lowerCamelCase !

ui/components/imageWorkspace/src/components/Features/TextInputFeature.svelte
```diff
@@ -43,12 +55,21 @@
   {#if isEditing}
     <Input
       value={textFeature.value}
-      on:change={(e) => onTextInputChange(e.currentTarget.value, textFeature.name)}
-      on:input={() => (isSaved = false)}
       type={inputType === "str" ? "text" : "number"}
       step={inputType === "int" ? "1" : "any"}
+      list="scene_availableValues_{textFeature.name}"
```
timothee-LJN, line 60: use lowerCamalCase here!

ui/components/imageWorkspace/src/components/Features/TextInputFeature.svelte
```diff
@@ -43,12 +55,21 @@
...
+      on:change={(e) => onTextInputChange(e.currentTarget.value, textFeature.name)}
+      on:input={() => (isSaved = false)}
       on:keyup={(e) => e.stopPropagation()}
     />
+    <datalist id="scene_availableValues_{textFeature.name}">
+      {#if "scene" in $itemFeaturesAvailableValues && textFeature.name in $itemFeaturesAvailableValues.scene}
```
timothee-LJN, line 66: I think we could narrow down `FeaturesAvailableValues` to something more explicitif 

```
interface FeaturesAvailableValues {
  scene: Record<string, Array<string>>;
  object: Record<string, Array<string>>;
}
```

that would make your component type safer and would avoid these type check everywhere. That would also make the code more readable. 
If the future we add more categories we will either refacto to something more generic or add the new category to the type.

cpvannier (changes requested):
Regarding the feature itself, currently this PR is only working for me on scene features. On object features, it only suggests me the current value after I hit enter, not the existing values.

Regarding the content of the PR, I think the TMP code in `dataset.py` and `ImageWorkspace.svelte` should be fixed here instead of in a future PR. I consider this feature to be a nice-to-have rather than a necessity, so I don't think we need to rush merging this into develop with temporary solutions.

The `categories` attribute of `DatasetInfo` should also be removed, along with its `DatasetCategory` class, as it is deprecated by `available_feat_values` and no longer needed.

The conversion from `InputFeaturesAvailableValues` to `FeaturesAvailableValues` in the frontend is a workaround that should be fixed as well. One solution could be to remove the `name` attribute from `DatasetAvailableFeatureValues` in the backend, and change the type of `available_feat_values` attribute in `DatasetInfo` to `dict[str, dict[str, DatasetAvailableFeatureValues]]` so that the feature names are already accessible as dictionary keys in the backend.

I also think the class and variable names are too long, I would rename to something like this:
- Backend
  - Rename `DatasetAvailableFeatureValues` class to `FeatureValues` and move it inside pixano/data/item/item_feature.py
  - Rename `available_feat_values` attribute of `DatasetInfo` to `feature_values` 
  - Rename `load_available_values` argument of `Dataset.load_info()` to `load_feature_values`
- Frontend
  - Remove `InputFeaturesAvailableValues` interface as mentioned above (have the correct format in the backend already)
  - Rename `FeaturesAvailableValues` interface to `FeatureValues`
  - Remove `inputFeaturesAvailableValues` and `itemFeaturesAvailableValues` variables as mentioned above (have `featureValues` in `DatasetInfo` instead)

ui/components/imageWorkspace/src/components/SaveShape/SaveShapeForm.svelte
```diff
@@ -76,6 +81,21 @@
 
       return [...oldObjects, ...(newObject ? [newObject] : [])];
     });
+
+    // add new inputs to lists of available values
+    if ("objects" in $itemFeaturesAvailableValues) {
```
timothee-LJN, line 86: this function looks the same as in `TextInputFeature.svelte`. Maybe you could abstract it in a function in an api file ?

ui/components/imageWorkspace/src/components/Features/TextInputFeature.svelte
```diff
@@ -43,12 +55,21 @@
...
-      on:change={(e) => onTextInputChange(e.currentTarget.value, textFeature.name)}
-      on:input={() => (isSaved = false)}
       type={inputType === "str" ? "text" : "number"}
       step={inputType === "int" ? "1" : "any"}
+      list="scene_availableValues_{textFeature.name}"
+      on:change={(e) => onTextInputChange(e.currentTarget.value, textFeature.name)}
+      on:input={() => (isSaved = false)}
       on:keyup={(e) => e.stopPropagation()}
     />
+    <datalist id="scene_availableValues_{textFeature.name}">
```
timothee-LJN, line 65: there is a `Combobox` component available (only available with search feature for now but could be easily improved). 
You can find documentation here : https://www.shadcn-svelte.com/docs/components/combobox 

I can help on that if you want?

ui/components/imageWorkspace/src/components/Features/TextInputFeature.svelte
```diff
@@ -43,12 +55,21 @@
...
+      {#if "scene" in $itemFeaturesAvailableValues && textFeature.name in $itemFeaturesAvailableValues.scene}
+        {#each $itemFeaturesAvailableValues.scene[textFeature.name].sort() as proposedValue}
```
timothee-LJN, line 67: that's cool to use reactive store, i never do it but will more in the future!

ui/components/imageWorkspace/src/ImageWorkspace.svelte
```diff
@@ -104,6 +109,30 @@
...
     }
   }
+
+  onMount(() => {
+    //TMP: should be removed when available value written in db.json, so present in DatasetInfo[] from getDatasets (S)
+    // but right now, we compute available feats only in getDataset (no S) route, used only for stats, and here
+    // later, we should get itemFeaturesAvailableValues directly from DatasetInfo["available_feat_values"]
+    api
+      .getDataset(currentDatasetId)
+      .then((datasetWithFeats) => {
```
timothee-LJN, line 119: I feel like this is not located in the most accurate component. Data fetching is mainly done in `src/routes/layout.svelte` . Maybe that could be something to reconsider but for now I think we should try to locate as many connexion to the API with in there, what do you think?

BertrandRenault:
I didn't remove "categories" as proposed yet, because without we can't export directly to COCO format.
Need to statute about the change request first

BertrandRenault:
> Regarding the feature itself, currently this PR is only working for me on scene features. On object features, it only suggests me the current value after I hit enter, not the existing values.

Auto-completion for object edition was omitted. Now added.

> Regarding the content of the PR, I think the TMP code in `dataset.py` and `ImageWorkspace.svelte` should be fixed here instead of in a future PR. I consider this feature to be a nice-to-have rather than a necessity, so I don't think we need to rush merging this into develop with temporary solutions.
>
> The `categories` attribute of `DatasetInfo` should also be removed, along with its `DatasetCategory` class, as it is deprecated by `available_feat_values` and no longer needed.

 
> The conversion from `InputFeaturesAvailableValues` to `FeaturesAvailableValues` in the frontend is a workaround that should be fixed as well. One solution could be to remove the `name` attribute from `DatasetAvailableFeatureValues` in the backend, and change the type of `available_feat_values` attribute in `DatasetInfo` to `dict[str, dict[str, DatasetAvailableFeatureValues]]` so that the feature names are already accessible as dictionary keys in the backend.

 
> I also think the class and variable names are too long, I would rename to something like this:
> 
> * Backend
>   
>   * Rename `DatasetAvailableFeatureValues` class to `FeatureValues` and move it inside pixano/data/item/item_feature.py
class removed, use of primitive type instead (dict[str, dict[str, list[string]]])

>   * Rename `available_feat_values` attribute of `DatasetInfo` to `feature_values`
Done

>   * Rename `load_available_values` argument of `Dataset.load_info()` to `load_feature_values`
Renamed to 'load_features_values'

> * Frontend
>   
>   * Remove `InputFeaturesAvailableValues` interface as mentioned above (have the correct format in the backend already)
Done

>   * Rename `FeaturesAvailableValues` interface to `FeatureValues`
Done
>   * Remove `inputFeaturesAvailableValues` and `itemFeaturesAvailableValues` variables as mentioned above (have `featureValues` in `DatasetInfo` instead)
Not yet

ui/components/imageWorkspace/src/ImageWorkspace.svelte
BertrandRenault: Done

cpvannier:
> I didn't remove "categories" as proposed yet, because without we can't export directly to COCO format. Need to statute about the change request first

True, this deserves its separate issue: pixano/pixano-project-manager#46

Also, regarding storing `feature_values` versus fetching the data with SQL requests, Jaonary suggested merging this PR as is with SQL requests only, and to look into adding a backend cache for storing data like this while the app is open, rather that writing it in the `db.json` files: pixano/pixano-project-manager#45

cpvannier (approved):

timothee-LJN (approved):

PR #135, opened 2024-02-12
chore: Freaze lockfile in CI

### LINKED ISSUE #97: Lock package version use in CI
## description
when package version update it can lead to bug. Locking version would avoid these bugs. Package version would have to upgraded manually.

timothee-LJN:
## issue
fixes #97 

## description
Instead of locking manually package version, I've added a flag `--frozen-lockfile` to GH actions. 
According the [docs](https://docs.npmjs.com/cli/v10/commands/npm-ci) this should have been the case since the command was run in CI already. So maybe there is something we did wrong or `CI=TRUE` was not explicit enough? 

`--frozen-lockfile` flag is [equivalent](https://stackoverflow.com/questions/70154568/pnpm-equivalent-command-for-npm-ci) to `npm ci` : https://docs.npmjs.com/cli/v10/commands/npm-ci

### DIFF
--- a/.github/workflows/format.yml
+++ b/.github/workflows/format.yml
@@ -49,7 +49,7 @@ jobs:
       - name: Install dependencies
         run: |
           cd ui
-          pnpm i
+          pnpm i --frozen-lockfile
 
       - name: Format frontend code with Prettier
         run: |
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -48,7 +48,7 @@ jobs:
       - name: Install dependencies
         run: |
           cd ui
-          pnpm i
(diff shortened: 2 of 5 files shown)


timothee-LJN:
@jrabary 
Checks are unsuccessful since `pnpm-lock.yaml` is in `gitignore` hence not commited. 
This does not seem to be a good practice:  
https://pnpm.io/git#:~:text=You%20should%20always%20commit%20the,able%20to%20skip%20package%20resolution

Is there any reason why it like so?

jrabary:
We can remove it from gitignore.

cpvannier (approved):
Typo in PR title (Freaze > Freeze)

PR #140, opened 2024-02-12
fix(front): add cancel option to no embeddings modal

### LINKED ISSUE #133: Blocked by modal if we don't have embeddings
since #123 (I suppose), we are stuck with modal
image
and the one asking for a model
if we have a dataset without embeddings

BertrandRenault:
close #133


timothee-LJN (approved):

cpvannier (commented):
After clicking cancel on the "No embeddings found" modal, clicking the button to close the smart tools list creates an error, and no other tools can be selected:
```
Uncaught (in promise) TypeError: ctx[3] is null
    update Rectangle.svelte:101
    update scheduler.js:119
    flush scheduler.js:79
```

BertrandRenault:
Right, but it's an issue independant from this PR. (even on develop, if you close smart tools list, as long as there is objects on screen, the error raises, because we then have "null" selectedTool. We should default to "PAN" when tools == null)
I open a new issue

cpvannier (approved):

PR #176, opened 2024-02-20
fix(front): Fix favicon 404 error

### LINKED ISSUE #149: Fix deployed app still looking for a favicon.ico
## Bug description

With the favicon now loaded with Svelte, when we launch the deployed app we get:
```bash
INFO:     127.0.0.1:40780 - "GET /immutable/assets/favicon.DroiEs1d.ico HTTP/1.1" 200 OK
INFO:     127.0.0.1:40748 - "GET /favicon.ico HTTP/1.1" 404 Not Found
```

Which means while Svelte loads the actual favicon, the HTML file is still looking for the default `favicon.ico`.
Maybe we can disable that to remove the warning?

## Steps to reproduce

1. Install the latest version of Pixano from PyPI (`pip install pixano==0.5.0b3`)
2. Launch the app (`pixano my_datasets/`)
3. Open the app and look at the logs in the terminal

## Expected behavior

The app should only load the favicon provided by the Svelte file and not look for a `favicon.ico` as well.

cpvannier:
## Issue

Fixes #149

## Description

Add empty favicon in HTML file to disable 404 error in deployed app. Favicon from Svelte file still loads correctly.

### DIFF
--- a/ui/apps/pixano/src/app.html
+++ b/ui/apps/pixano/src/app.html
@@ -3,6 +3,7 @@
   <head>
     <meta charset="utf-8" />
     <meta name="viewport" content="width=device-width, initial-scale=1" />
+    <link rel="icon" href="data:," />
     %sveltekit.head%
   </head>
   <body data-sveltekit-preload-data="hover">


ui/apps/pixano/src/app.html
```diff
@@ -3,6 +3,7 @@
   <head>
     <meta charset="utf-8" />
     <meta name="viewport" content="width=device-width, initial-scale=1" />
+    <link rel="icon" href="data:," />
```
timothee-LJN, line 6: question: what is also an error if you delete this line?

timothee-LJN (approved):

ui/apps/pixano/src/app.html
cpvannier: The error without this line is a `"GET /favicon.ico HTTP/1.1" 404 Not Found` in the deployed app.

I am guessing that without a specified icon, the HTML file will automatically look for a "favicon.ico" in the current folder.

The Svelte favicon still loads correctly with our without this line, this just removes the useless 404 message.

BertrandRenault (approved):

PR #181, opened 2024-02-21
feat: docker build files & doc

### LINKED ISSUE #167: Create a docker image of pixano app

BertrandRenault:
## Issue

Close #167 

## Description

docker directory with Dockerfile, build and run scripts, and README.md on how to run Pixano Docker image, and build it if wanted.

- pixano.sh script will pull image from DockerHub if not present, and run Pixano
- build.sh allows to build a local image on your system (instead of pulling from DockerHub)
- also, allowed user can uncomment the "docker push" line in build.sh, to push a new version on DockerHub


cpvannier (changes requested):
From what I have seen on other repositories should be on the root directory of the project repository.

- With the Dockerfile on the root directory, I imagine the `.dockerignore` will not be needed anymore.

- The `README.md` file could be turned into a page of the documentation website, with the `docker/assets` moved in `docs/assets/docker`.

- The `build.sh` script seems very redundant, as it seems to only convert its `$1` `$2` `$3` args into the Dockerfile `$tag`, `$pixano_version` and `$pixano_inference_version`.
Why use that script instead of calling the Dockerfile directly with `docker build`? 
To simplify the `docker build` command line, I would argue that we could remove the `$pixano_version` and `$pixano_inference_version` args altogether, and just have `pip install` fetch the latest versions.
The args were useful while working with beta versions, as those were not the default choices for pip, but this it not an issue anymore. 

- Finally, I think we should remove the `pixano.sh` script as well if we can. With clear enough documentation, I think it is better to ask the user to run `docker pull` and `docker run` themselves, rather than running a script that will just add some default values for the parameters like `build.sh`.
Also users looking to run Pixano through Docker rather than through PyPI probably already have experience with Docker anyway, so it would be better to document how they should use the docker commands with Pixano directly rather than hiding them in scripts.

BertrandRenault:
> * From what I have seen on other repositories, the `Dockerfile` should be placed on the root directory of the project repository.

Here, Dockerfile doesn't use the local pixano install (pip install . ), but pip install pixano, from PyPi. In fact the best approch would be a separate repo for the docker directory.

> * With the `Dockerfile` on the root directory, I imagine the `.dockerignore` will not be needed anymore.

I'll remove the .dockerignore, useless here as there is no COPY line anymore.

> * The `README.md` file could moved to the documentation website as a subsection in `docs/getting_started/installing_pixano.md`, with the `docker/assets` moved in `docs/assets/docker`.

It could.

> * The `build.sh` script seems very redundant, as it seems to only convert its `$1` `$2` `$3` args into the Dockerfile `$tag`, `$pixano_version` and `$pixano_inference_version`.
>   Why use that script instead of calling the Dockerfile directly with `docker build`?
>   To simplify the `docker build` command line, I would argue that we could remove the `$pixano_version` and `$pixano_inference_version` args altogether, and just have `pip install` fetch the latest versions.
>   The args were useful while working with beta versions, as those were not the default choices for pip, but this it not an issue anymore.

I agree for version, as we have now 0.5.0. Thought I will let the ability to choose a version.

> * Finally, I think we should remove the `pixano.sh` script as well if we can. With clear enough documentation, I think it is better to ask the user to run `docker pull` and `docker run` themselves, rather than running a script that will just add some default values for the parameters like `build.sh`.
>   Also users looking to run Pixano through Docker rather than through PyPI probably already have experience with Docker anyway, so it would be better to document how they should use the docker commands with Pixano directly rather than hiding them in scripts.

All this Docker thing was thought as a very user-friendly, easy to use (Docker Desktop). Just run one line and you have a pixano container running. It has not been thought for experienced Docker users. Experienced users can look at the scripts if they want to tweak something.

timothee-LJN (approved):

juliendenize:
Close because done in #340 #339

PR #220, opened 2024-03-28
feat(front): Add mask annotations on video frames

### LINKED ISSUE #214: Mask annotation in video frame

timothee-LJN:
## Issue

Fixes #214 

## Description

- This branch is pulled from #185
- To add mask to video, the data model had to evolve compared to #185 since `tracklets` need to accept both `bbox`or `mask` 
- A `tracklet` is now an array of `KeyVideoFrame`, each can have a `bbox` or a `mask` 
- I would advise to check first `datasetTypes.ts` file to understand the data structure
- The mask linear interpolation does not feel super snappy since we need to convert `mask.count` to `svg` and then to coordinates and then apply the `simplifyJs` API on each calculation. We also refer on array index to retrieve points, which is not working perfectly (causing sometimes the mask to turn around). Maybe this data structure should be given a second though?


BertrandRenault:
Closed as irrevelant after data model change
Could still be used as reference / cherry pick when we will add mask for video

PR #230, opened 2024-04-15
fix(front): look for input rect in drag-rect instead of input

### LINKED ISSUE #229: Rectangle input with SAM model doesn't work anymore since 0.5.1
## Bug description

With Pixano 0.5.1, when we use the rectangle input for smart segmentation with SAM, the resulting mask is totally wrong. it seems the input is not correct
Most probably related to #205 

## Steps to reproduce

1. With pixano 0.5.1, open any dataset with SAM embeddings
2. Choose rectangle input for interactive segmentation
3. Draw a rectangle on any object
4. we have no mask, or a very wrong one

## Expected behavior

The correct mask.

## Additional context

This appears with pixano 0.5.1 (pixano 0.5.0 is OK in this regard), hence the link to #205

BertrandRenault:
## Issue

Fixes #229 

## Description

input rectangle for segmentation is not created under "input" group anymore (since #205 )
So we find it by it's name "drag-rect" now

### DIFF
--- a/ui/components/canvas2d/src/Canvas2D.svelte
+++ b/ui/components/canvas2d/src/Canvas2D.svelte
@@ -679,21 +679,17 @@
     //get box as Box
     let box: Box = null;
     const viewLayer: Konva.Layer = stage.findOne(`#${viewId}`);
-    const inputGroup: Konva.Group = viewLayer.findOne("#input");
-    for (const rect of inputGroup.children) {
-      if (rect instanceof Konva.Rect) {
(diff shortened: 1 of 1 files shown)


BertrandRenault:
@timothee-LJN I requested your review on this because you made the changes for #205, so if you think there may be something I missed there, let me know.

By the way, I tested it and it's working with theses changes

juliendenize (approved):

timothee-LJN (approved):

PR #250, opened 2024-05-27
feat(front): Add image adjustment tools

### LINKED ISSUE #241: Image adjustement tools
Image 8bits/16bits
- [ ] Adjust contrast
- [ ] Adjust brightness
- [ ] Histogram level

ulysse-bonneau:
## Issue

Fixes #241

## Description

For 8-bit and 16-bit images :
- [x] Adjust contrast
- [x] Adjust brightness
- [x] Histogram level

Filters are disabled for videos.

image


timothee-LJN:
I've just tested and videos don't play right, the zoom level change everyframe.

Also, sliders are not centered 

image

ulysse-bonneau:
> I've just tested and videos don't play right, the zoom level change everyframe.
> 
> Also, sliders are not centered on chromium / mac
> image

The video playing should be smoother now. I haven't looked into the sliders issue yet.

ui/components/datasetItemWorkspace/src/components/DatasetItemViewer/ImageViewer.svelte
```diff
@@ -27,56 +32,122 @@
...
+   */
+  const loadImages = async (views: Record<string, ItemView>): Promise<ImagesPerView> => {
+    const images: ImagesPerView = {};
+    const promises: Promise<void>[] = Object.entries(views).map(async ([key, value]) => {
+      const img: ImageJS = await ImageJS.load(`/${value.uri}`);
+
+      $itemMetas.color = img.channels === 1 ? "grayscale" : "rgba";
+      $itemMetas.format = (img.bitDepth as number) === 8 ? "8bit" : "16bit";
+
+      if ($itemMetas.format === "16bit" && $itemMetas.color === "grayscale") {
```
BertrandRenault, line 83: Is it a requirement to normalize only non 8bit grayscale ? 
- test is not 16bit, it's non 8bit (from [ImageJS doc](https://github.com/image-js/image-js/blob/9ab86a86f6c13a9a7d14c62566c1396c3c6f54f4/src/image/Image.js#L36)) it could be 1, 8 or 16 (maybe 32 also, not clear)
- normalize only grayscale ?

ulysse-bonneau: I don't know if normalize is the right term for the function, its purpose is to let us convert the image from 16-bit to 8-bit ourselves, rather than use the default HTML/Konva conversion. It also allows us to select the range of pixel values we want to display. We could technically add support for 8-bit images here, but I don't know if that is the right place to do so, as we could also use a custom Konva filter in Canvas2D together with the other filters. About grayscale, I only implemented the support for it because Jaonary told me that the 16-bit images of the user would be grayscale. It is possible however to easily add support for RGB(A) 16-bit images.

BertrandRenault (approved):

PR #260, opened 2024-05-30
feat(back): save item + video: Tracklet table handle

BertrandRenault:
## Issue

Save - back side
Read video - back side : tracklet table

## Description

- [x] handle tracklet table for video
- [x] update item features
- [x] add new objects
- [x] update existing objects
- [x] delete removed objects

### DIFF
--- a/pixano/datasets/dataset.py
+++ b/pixano/datasets/dataset.py
@@ -18,13 +18,11 @@ import duckdb
 import lancedb
 from lancedb.pydantic import LanceModel
 from lancedb.query import LanceQueryBuilder
-import pyarrow as pa
 from pydantic import ConfigDict
 from s3path import S3Path
 
 from .dataset_features_values import DatasetFeaturesValues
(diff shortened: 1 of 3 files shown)


pixano/datasets/dataset.py
```diff
@@ -688,6 +691,111 @@ def get_embedding(
         """
         return self.get_embedding(idx, 1, select)[0]
 
+    def save_item(
```
juliendenize, line 694: I do not think that save_item should handle Front objects. IMO the front should only communicate with the REST API which should handle the conversion for the back.

The back should only handle lance models so that DS can provide in a pythonic way new items to save.

juliendenize (changes requested):

pixano/datasets/dataset.py
BertrandRenault: Right, I moved this function to REST api

juliendenize (approved):

PR #285, opened 2024-06-21
feat(front): Improve Table component scrolling

ulysse-bonneau:
## Issue

Fixes #31 

## Description

The Table component now scrolls with a sticky header for column names, rather than scroll the whole page.
Having the Table component fixed and scrollable allows the page navigation buttons below to be always accessible at a fixed point. This allows moving through large datasets quicker by clicking repeatedly on next or previous page buttons without having to scroll down each time.
image

### DIFF
--- a/ui/components/table/src/Table.svelte
+++ b/ui/components/table/src/Table.svelte
@@ -27,12 +27,11 @@
   import { createEventDispatcher } from "svelte";
   import { readable } from "svelte/store";
   import SortableList from "svelte-sortable-list";
-  import { createTable, Subscribe, Render, createRender } from "svelte-headless-table";
+  import { createTable, Subscribe, Render, DataBodyCell, type RenderConfig } from "svelte-headless-table";
   import { addColumnOrder, addHiddenColumns } from "svelte-headless-table/plugins";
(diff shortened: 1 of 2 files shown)


ui/components/table/src/Table.svelte
```diff
@@ -27,12 +27,11 @@
   import { createEventDispatcher } from "svelte";
   import { readable } from "svelte/store";
   import SortableList from "svelte-sortable-list";
-  import { createTable, Subscribe, Render, createRender } from "svelte-headless-table";
+  import { createTable, Subscribe, Render, DataBodyCell, type RenderConfig } from "svelte-headless-table";
```
BertrandRenault, line 30: this new imports seems not used

BertrandRenault (approved):
I tested, it works fine except that if we change column settings (visibilty, order) they're reset on page change.
But in fact it was already the case before this change. 
I opened a separate issue for this #286

ui/components/table/src/Table.svelte
ulysse-bonneau: Thank you, I removed it.

PR #289, opened 2024-06-27
TMP single-view+prompt-chooser

BertrandRenault:
## Issue

TEMPORARY
There is some issues with multiview
So we will temporarly show view separatly. User is prompted to select which view to display

## Description

By showing only one view at a time, we get rid of multiview issues, until we finished to fix them

- back: filter elements (views/objetcts/...) for the selected view
- front: user prompt view selection


BertrandRenault:
Closed as irrevelant now after changes in data model
(Note: the front code could still be used in case of)

PR #297, opened 2024-07-04
Test types schemas

juliendenize:
## Description

- Add tests for schemas and types in the backend
- rename is_obj_of_type to is_subclass_strict
- add BaseType for Pixano types and _PIXANO_TYPES_REGISTRY
- add model validations for various types => to be tested in future PR


pixano/datasets/features/schemas/sequence_frame.py
```diff
@@ -57,11 +59,12 @@ def create_sequence_frame(
     image = create_image(item_id, url, id, width, height, format, other_path)
     return SequenceFrame(
         id=image.id,
+        item_id=item_id,
+        sequence_id=sequence_id,
```
BertrandRenault, line 63: Is sequence_id really a required field ?

BertrandRenault (approved):

pixano/datasets/features/schemas/sequence_frame.py
juliendenize: It seems to be the case as it is how we link multiple sequence frames to one video.

PR #311, opened 2024-07-22
limit objects features to top level entity

BertrandRenault:
## Issue

Right now, front can't manage "dynamic" metadata (video).
For now, we manage only static metadata (track level)

## Description

Do not send annotation and sub-entities metadata to front

### DIFF
--- a/pixano/app/api/items.py
+++ b/pixano/app/api/items.py
@@ -338,7 +338,7 @@ async def get_dataset_item(  # noqa: D417
     NoneMask = CompressedRLE.none()
     NoneKeypoints = KeyPoints.none()
 
-    def findTopEntity(ann: Annotation) -> Entity | None:
+    def find_top_entity(ann: Annotation) -> Entity | None:
         """Find top entity of an annotation.
 
         Args:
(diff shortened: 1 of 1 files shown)


pixano/app/api/items.py
```diff
@@ -338,28 +338,40 @@ async def get_dataset_item(  # noqa: D417
...
+
+        Returns:
+            Entity | None: top Entity for this annotation
+        """
+        try:
+            return findTopParentEntity(ann.entity)
+        except ValueError:
+            return None
+
+    def findTopParentEntity(entity: Entity) -> Entity | None:
```
juliendenize, line 355: This is not how we should name functions in python => find_top_parent_entity however i've seen that other part of the API have this kind of naming functions.

Should we adress them now or when we will refactor the front and API  ?

pixano/app/api/items.py
```diff
@@ -338,28 +338,40 @@ async def get_dataset_item(  # noqa: D417
...
-    def findEntity(ann: Annotation) -> Entity | None:
-        for group_entity in groups[_SchemaGroup.ENTITY]:
-            entity = next(
-                (entity for entity in getattr(item, group_entity) if entity.id == ann.entity_ref.id),
-                None,
-            )
-            if entity:
-                return entity
-        return None
+    def findTopEntity(ann: Annotation) -> Entity | None:
```
juliendenize, line 341: Same comment as below

pixano/app/api/items.py
BertrandRenault: you right, corrected

PR #320, opened 2024-07-30
tracking: data opening optimization

### LINKED ISSUE #313: Tracking : Slow tracking dataset opening with explorer
Opening a large tracking dataset with pixano explorer takes a lot of time. It seems that the entire dataset needs to be loaded at once before being shown to the user. This makes it unusable for annotating more than a few seconds.

glapouge:
## Issue

Tracking dataset would take very long to load.

Fixes #313 
Changed the data loading paradigm where a buffer of correspondence is first created and then used to load all data fater.

### DIFF
--- a/pixano/app/api/items.py
+++ b/pixano/app/api/items.py
@@ -411,12 +411,10 @@ async def get_dataset_item(  # noqa: D417
                 obj["features"] = features
                 objects.append(obj)
     else:  # video
-        trackid2trackletslist = defaultdict(list) #trackid2trackletslist
-        trackid2entityidlist = defaultdict(list) #trackid2entityidlist
+        trackid2trackletslist = defaultdict(list)  # trackid2trackletslist
+        trackid2entityidlist = defaultdict(list)  # trackid2entityidlist
         entity_id2trackid = {}
-        trackid2annotationidlist = {annotation_group:defaultdict(list) for annotation_group in groups[_SchemaGroup.ANNOTATION]} #trackid2entityidlist
         trackid2track = {}
-        verbose = False
 
         ## Store a cache of references for faster loading
         # gather tracklets for each track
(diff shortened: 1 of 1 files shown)


glapouge:
My first pull request, please review :)

BertrandRenault:
minor fixes:
- ruff format
- type hinting for mypy
- removed unused vars
- fix a missing data in some case

pixano/app/api/items.py
```diff
@@ -474,8 +472,7 @@ def find_top_parent_entity(entity: Entity) -> Entity | None:
                 kept_track_ids.append(track_id)
```
BertrandRenault, line 472: kept_track_ids seems useless right now, but it may become usefull later, so let's keep it

BertrandRenault (approved):

pixano/app/api/items.py
```diff
@@ -411,107 +411,118 @@ def find_top_parent_entity(entity: Entity) -> Entity | None:
                 obj["features"] = features
                 objects.append(obj)
     else:  # video
-        tracks = defaultdict(list)
-        entity_id = defaultdict(list)
+        trackid2trackletslist = defaultdict(list)  # trackid2trackletslist
```
juliendenize, line 414: Can we do some refactoring so that the variable names ae more easily readable and consistent  ?
trackid2trackletslist => track_id_to_tracklet_list
for instance

pixano/app/api/items.py
```diff
@@ -411,107 +411,118 @@ def find_top_parent_entity(entity: Entity) -> Entity | None:
...
+                if is_tracklet(type(annotation)):
+                    continue
+                entity_id = annotation.entity_ref.id
+                track_id = entity_id2trackid[entity_id]
+                frame_index = next(
+                    (
+                        view["frame_index"]
+                        for view in views[annotation.view_ref.name]
+                        if view["id"] == annotation.view_ref.id
+                    ),
```
juliendenize, line 462: we suppose that the view is sequene frame which is ok for now but we should keep in mind that later we will support videos.

pixano/app/api/items.py
juliendenize: yes everything will be refactor at some moint anyway

juliendenize (changes requested):

PR #331, opened 2024-09-23
Add Source and timestamps

juliendenize:
## Description

Add Source group to add information for data provenance.

Add created_at, updated_at fields for every tables.


pixano/app/models/base_schema.py
```diff
@@ -29,16 +31,56 @@ class BaseSchemaModel(BaseModel, Generic[T]):
...
+        `self` is explicitly positional-only to allow `self` as a field name.
+
+        Args:
+            created_at: The creation date of the object.
+            updated_at: The last modification date of the object.
+            data: The data of the object validated by Pydantic.
+        """
+        if created_at is None or updated_at is None:
+            if updated_at is not None or created_at is not None:
+                raise ValidationError("Both 'created_at' and 'updated_at' should be set.")
```
BertrandRenault: I suggest we allow a None "updated_at" with a non None "created_at", as at creation time, updated_at == created_at
Badly it means changes to some tests and inits...

pixano/app/models/base_schema.py
```diff
@@ -29,16 +30,56 @@ class BaseSchemaModel(BaseModel, Generic[T]):
...
+        `self` is explicitly positional-only to allow `self` as a field name.
+
+        Args:
+            created_at: The creation date of the object.
+            updated_at: The last modification date of the object.
+            data: The data of the object validated by Pydantic.
+        """
+        if created_at is None or updated_at is None:
+            if updated_at is not None or created_at is not None:
+                raise ValueError("Both 'created_at' and 'updated_at' should be set.")
```
BertrandRenault, line 53: I suggest we allow a None "updated_at" with a non None "created_at", as at creation time, updated_at == created_at
Badly it means changes to some tests and inits...

juliendenize: It should  be better now

PR #333, opened 2024-09-30
Add integrity checks

juliendenize:
## Description

Add integrity checks for:
- analysing a dataset
- analysing a table
- adding or updating data

Integrity checks can be performed when building the dataset.


BertrandRenault (commented):
Why do you check before add/update ? Wouldn't it be better to check after any modification ?

juliendenize:
> Why do you check before add/update ? Wouldn't it be better to check after any modification ?

Because then the modification already happens which is not intended to keep dataset consistency.

BertrandRenault (approved):

PR #334, opened 2024-10-04
Add source to builder

juliendenize:
## Description

Add source handler in builders

Add OTHER group in SourceKind


pixano/datasets/builders/folders/base.py
```diff
@@ -115,6 +116,7 @@ def generate_data(
         Returns:
             An iterator over the data following data schema.
         """
+        source_id = None
```
BertrandRenault, line 119: This is temporary while we don't manage other sources yet ?

BertrandRenault (approved):

pixano/datasets/builders/folders/base.py
juliendenize: In line 151 I check if source is None:
- yes: create a source
- no: do nothing as the source is already created.

PR #358, opened 2024-11-08
add a frame buffer for faster play

BertrandRenault:
## Issue

Images were loaded on demand. This could slow down the play of video

## Description

Asynchronously, load and store each frame as HTMLImageElement (almost ready to display), populating the buffer frame by frame (across views)

Note: 
pro: faster play when frame loaded, populating asynchronously so very little overhead cost.
con: all calls to API for views at the start, all frames are loaded even if we just want a quick view on the item.

### DIFF
--- a/ui/components/datasetItemWorkspace/src/components/DatasetItemViewer/VideoViewer.svelte
+++ b/ui/components/datasetItemWorkspace/src/components/DatasetItemViewer/VideoViewer.svelte
@@ -125,8 +125,9 @@ License: CECILL-C
   let expanding = false;
 
   let imagesPerView: ImagesPerView = {};
+  const imagesPerViewBuffer: Record<string, { id: string; element: HTMLImageElement }[]> = {};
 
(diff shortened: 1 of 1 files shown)


BertrandRenault:
> Also, the buffer should load frame by frame, not view by view: here, it will load the whole 10 minutes for camera 1, then the whole 10 minutes for camera 2, etc. So when playing, camera 6 will not be displayed at all until all the other cameras have fully loaded.

You right, it should load each view of frame 0 then each view of next frame etc. ,  but the array of promise is wrongly ordered it seems. I will try to reorder this

BertrandRenault:
> This is a bit of a problem actually. I am currently testing Pixano for multi-view video datasets (5 to 6 cameras, 10 minutes chunks), and v0.6.0b5 works fine, but with this commit, the tab often ends up crashing because it tries to load too many images at once.

Ok I will try to load a fixed number of frames in advance

PR #365, opened 2024-11-26
Separate bounding box xywh conversion from denormalization in frontend

### LINKED ISSUE #364: Bounding boxes not converted from xyxy to xywh in frontend
## Bug description

Conversion of xyxy to xywh coordinates to display bounding boxes in the frontend is only triggered if the coordinates also need to be denormalized.

## Steps to reproduce

Opening a dataset with bounding boxes stored as xyxy denormalized coordinates.

## Expected behavior

Conversion from xyxy to xywh should be triggered regardless if coordinates are normalized or not.

cpvannier:
xyxy coordinates were not converted to xywh to display in the frontend if they were already denormalized

## Issue

Fixes #364

## Description

Separate the bounding box xyxy to xywh conversion step from the denormalization step in frontend


BertrandRenault (approved):
Careful, there is still some weird behaviour with xyxy coords.
for instance if we edit a bbox (change coords)
(https://github.com/pixano/pixano-project-manager/issues/83)
Also, we should put format back to xyxy after ui conversion (to match user format). It's not done actually (should/may be done in DatasetItemWorkspace.svelte > front2back)

PR #368, opened 2024-11-27
circular buffer

### LINKED ISSUE #366: Crashes when buffering too many frames
## Bug description

> add a frame buffer for faster play (@BertrandRenault, #358)
> con: all calls to API for views at the start, all frames are loaded even if we just want a quick view on the item.

This is a bit of a problem actually. I am currently testing Pixano for multi-view video datasets, and the v0.6.0b5 works fine, but when running on the main branch with this commit, the tab often ends up crashing because it tries to load too many frames at once.

## Steps to reproduce

Opening a dataset item with long videos (e.g. 10 minutes or more) and/or multiple views (e.g. 5 cameras or more)

## Expected behavior

The buffer is a great idea, the stutter when playing in v0.6.0b5 is fixed in this commit (when it doesn't crash), but ideally it should only load the next few seconds, not the whole video.

Also, the buffer should load frame by frame, not view by view: here, it will load the whole 10 minutes for camera 1, then the whole 10 minutes for camera 2, etc. So when playing, camera 6 will not be displayed at all until all the other cameras have fully loaded.

## Additional context

_Originally posted by @cpvannier in https://github.com/pixano/pixano/issues/358#issuecomment-2500897742_

BertrandRenault:
## Issue

Pixano app do not have a max size for media buffer, leading to crash (memory full)

Fixes #366 

## Description

Implement a circular buffer (actually 50 frames after and 5 frames before current frame)
-- this settings are currently hard coded

tech detail: the 50 next (including current) and 5 previous are loaded at start, and each time we change current frame index. If any of the required buffer frames are not present, load it. frmaes outside of this range are removed from buffer, so it always have a size of 55 frames ( * nb views) --- please note that we add before deleting, so at max we can have 55 * 2 images in the buffer. -- in usual cases (play or step forward/backward), only one frame is added / removed each time

### DIFF
--- a/ui/components/datasetItemWorkspace/src/components/DatasetItemViewer/VideoViewer.svelte
+++ b/ui/components/datasetItemWorkspace/src/components/DatasetItemViewer/VideoViewer.svelte
@@ -55,6 +55,8 @@ License: CECILL-C
   export let embeddings: Record<string, ort.Tensor>;
   export let currentAnn: InteractiveImageSegmenterOutput | null = null;
 
+  const IMAGE_BUFFER_SIZE = 10;
+
   $: {
     if (selectedItem) {
       currentFrameIndex.set(0);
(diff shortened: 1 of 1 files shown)


juliendenize:
In the related issue I wrote a comment about how I see things, this PR solves it in a different way. I don't mind not going my way however I would still like a discussion on why we're going for a fixed number per view and not a global multi-view storage.

If it is a question of time and having a quick fix to prevent @cpvannier to have a problem, I can get that but i'm not convinced it fixes things correctly. Here you suppose the network is not an issue which might not be the case for everyone. There will be intensive query to the DB even for small videos.

BertrandRenault:
It is of course possible to have a fixed number of images in buffer, independantly of number of views. (or even a fixed size depending also on image size). We just need to divide the fixed size by number of view. But in my opinion what really matter is that we can play smoothly (with play & step forward/backward) so having a few step backward and forward seems appropriate.
A downside versus previous "complete" buffer is that moving randomly in the video (ie. sliding through the player) is not as smooth as before (but obviously we can't keep this complete buffer)

Regarding bandwith... Not sure we could really control this. We are trying to read synchronously (N views) video at same time... It requires a good network indeed. Maybe we could use some compressed format, or even convert sequence frames to video to use some existing optimized player ?

We should check the bottlenecks. Annotations also can be quite heavy, and are all loaded at once (DatasetItem)

BertrandRenault:
added a timeout to prevent loading for outdated updates (it happens if we slide quickly the player). It widely reduce the amount of network traffic
Also, buffer is now of fixed size. We can set this size, and the ratio of backward frame in the buffer (it's still in number of images, regardless of  real image size). So the more views we have, the smaller the buffered range will be.

cpvannier:
> If it is a question of time and having a quick fix to prevent @cpvannier to have a problem, I can get that but i'm not convinced it fixes things correctly.

The buffer that was pushed in v0.6.0b6 by #358 is definitely not viable for long videos as explained in #366, since it tries to load all frames into memory, and does it view by view all the way to the end of the video, instead of frame by frame.

So an update is definitely needed for the next beta release, but I'd rather we find a stable solution that a quick fix.

> Here you suppose the network is not an issue which might not be the case for everyone. There will be intensive query to the DB even for small videos.

I also think that 200 frames per view can quickly become too intensive. Maybe 20-50 frames per view is enough.

> https://github.com/pixano/pixano/issues/366#issuecomment-2503441223
> Because having a few seconds for 300fps video or 24 is not the same. The buffer should probably be of a fixed-size in terms of RAM storage and it should be multiview. This way, we handle the case where we have higher or lower framerates but also handle the multiview setting. The more views you have the less cache per view you can handle.

I agree that the framerate and the number of views will impact the buffer. However:
- Regarding the framerate: Right now the player has no notion of "fps". So the buffer proposed by Bertrand is defined with a number of frames, rather that a number of seconds. So for 300 fps of 24 fps videos, it will be the same.
- Regarding the number of views: I don't know if the buffer should be multi-view. If we find that, for example, loading the next 50 frames per view is the best choice for smooth playback:
  - Either we assume a low number of views, like 4 views, and set the buffer to 200 frames. But then, for a dataset with 8 views, that's only 25 frames per view, half of what we decided was best, and that may impact the playback for the video.
  - Or we assume a high number of views, like 10 views, and set the buffer 500 frames. But for a single-view dataset, it seems unnecessary to load that much.
  - So maybe keeping the buffer size per view is best?

BertrandRenault:
> I also think that 200 frames per view can quickly become too intensive. Maybe 20-50 frames per view is enough.
With latest changes, the frame_buffer is set to 200 images. so with 4 views it will be 50 images per view

We could eventually remove the view factor (so buffer_size depends on nomber of view, like it was just before the last commit (https://github.com/pixano/pixano/pull/368/commits/7e6651fcfae7dd15fbec31d302be555964ce7c41))

Another option is to make the buffer size a dataset settings (user defined to overwrite default), given in DatasetBuilder and stored in info.json

Or to complety rethink this. But you should try it now, it's working quite smooothly in my opinion (of course it's not a good idea to play a multiview video over a low bandwith network...)

juliendenize:
I think I will merge this one if you agree @cpvannier that in some extent fixes things for you, we plan on changing how multiview works later so we will have to go through this again probably and improve over this PR.

cpvannier:
Yes, you can merge this. :+1: 
This is definitely more efficient than the current buffer, so it is a better placeholder solution until you come back to it later.

cpvannier (approved):

PR #369, opened 2024-11-27
thumbnails in browser

BertrandRenault:
## Issue

In Pixano browser, the thumbnails were in fact base64 of the full image.
For a dataset with several views and items, and high res images, it's quite huge to display thumbnail (thumbnailing was done front size)

## Description

Images are now shrinked to thumbnail (128 * 128) back size, before convertion to base64
Update to lancedb 0.16.0


cpvannier (approved):
Thanks!

On the subject of thumbnails, maybe you can also use this PR to remove `pixano.dataset.utils.image.py`? It only contains the `image_to_thumbnail()` function that served a similar purpose a few months ago, but is now unused.

BertrandRenault:
you right, I wanted to use it but it's not convenient as it requires to convert (PIL/bytes) even more

pixano/features/schemas/views/image.py
```diff
@@ -39,31 +39,43 @@ class Image(View):
     format: str
 
     @overload
-    def open(self, media_dir: Path, output_type: Literal["base64"]) -> str: ...
+    def open(self, media_dir: Path, output_type: Literal["base64"], thumbnail_size: tuple[int, int] | None) -> str: ...
     @overload
-    def open(self, media_dir: Path, output_type: Literal["image"]) -> PILImage.Image: ...
-    def open(self, media_dir: Path, output_type: Literal["base64", "image"] = "base64") -> str | PILImage.Image:
+    def open(
+        self, media_dir: Path, output_type: Literal["image"], thumbnail_size: tuple[int, int] | None
```
juliendenize, line 45: The default value should be None

juliendenize (changes requested):
Could you rename thumbnail_size variables to size or resize ? It can be use for other things than thumbnails.

pixano/features/schemas/views/image.py
juliendenize: Yes my bad I overlooked for the default value.

Concerning the thumbnails, I better see the case now, however I feel it might be the wrong place to generate them. Indeed, it is more related to image transformation than to images.

Do you agree ? If so should we move the logic of creating thumbnail (and even base64) in an utils files (maybe the one deleted) and only keep PIL image opening inside the image ?

BertrandRenault: Yes I agree. I put it inside open_url because at first the removed image_to_thumbnail required a little too long chain of format convert, as it was wrote, so I found it more convenient to put it there.
I could definitively write a more usable to_thumbnail(image: PIL.Image, output_type=Literal["base64", "image"]). 
I suggest to put it in features/utils/image.py instead of the removed one. Except from the removed function, all images utils are there.

PR #370, opened 2024-11-28
fix(front): Fix getCurrentImageTime()

cpvannier:
## Description

Previous implementation of getCurrentImageTime() displays both "0:60" then "1:00" when passing from one minute to the next


BertrandRenault (approved):

cpvannier:
Yes, I am planning to create an issue / PR about this, and more generally the use of the timestamps stored in the SequenceFrame data, for a more accurate time representation in the VideoViewer.

I found this bug in getCurrentImageTime while working on that.

PR #376, opened 2024-12-03
fixed tracklet size

### LINKED ISSUE #378: VideoInspector - Right-click option does not work if not used on current frame
## Bug description

In the VideoInspector (the timeline displaying tracklets and bounding boxes), when right-clicking on a tracklet, you have a few options, like "Add a point" if it is missing a bounding box for a specific frame, "Split tracklet", and "Delete tracklet".

When selecting "Add a point", if the relevant frame is not the current frame displayed by the player, it will run into an error. As it tries to move to the specified frame to create the bounding box, the current time will actually turn into "NaN:NaN.NaN" and reset to the beginning of the video.

This is because right-clicking should already move the cursor to the selected frame before selecting "Add a point", but the invisible button that does this is actually located below the tracklet, and so it doesn't get trigger if you click on the tracklet itself, only if you click a few pixels below it. 

The same issue also seems to affect "Split tracklet", although the cursor will not reset to NaN, the tracklet just won't be split.

## Steps to reproduce

1. Open a video dataset item
2. Right click on a tracklet frame that is not the current frame
3. Click "Add a point" or "Split tracklet"

## Expected behavior

The invisible button in charge of moving the cursor to the frame where the user clicked (either left click or right click) should be centered on the tracklet. This would fix the issue.

As a fail-safe, it could also be an option to trigger the function that moves the cursor to the selected frame when selecting a right-click option like "Add a point" or "Split tracklet", just in case it wasn't already triggered by the right click.

BertrandRenault:
## Issue

At 5 or more views, the tracklet size is too small (track size / n views) in videoInspector to interact with (and currentFrameIndex is not set also because of a reactive component being disabled because too small, which leaded to several bugs)

Fix #378 

## Description

The tracklet now have a fixed size, and the track have a variable size depending on number of views.

### DIFF
--- a/ui/components/datasetItemWorkspace/src/components/VideoPlayer/ObjectTrack.svelte
+++ b/ui/components/datasetItemWorkspace/src/components/VideoPlayer/ObjectTrack.svelte
@@ -311,9 +311,9 @@ License: CECILL-C
 
 {#if track && tracklets}
   <div
-    class="flex gap-5 relative h-12 my-auto z-20"
+    class="flex gap-5 relative my-auto z-20"
     id={`video-object-${track.id}`}
-    style={`width: ${$videoControls.zoomLevel[0]}%`}
+    style={`width: ${$videoControls.zoomLevel[0]}%; height: ${Object.keys(views).length * 20}px;`}
     bind:this={objectTimeTrack}
     role="complementary"
   >
--- a/ui/components/datasetItemWorkspace/src/components/VideoPlayer/ObjectTrack.svelte
+++ b/ui/components/datasetItemWorkspace/src/components/VideoPlayer/ObjectTrack.svelte
@@ -90,123 +90,117 @@ License: CECILL-C
     );
   };
 
-  const onAddKeyItemClick = (tracklet: Tracklet | MouseEvent) => {
-    if (tracklet instanceof MouseEvent) {
-      //MouseEvent, need to determine view_id
-      //TODO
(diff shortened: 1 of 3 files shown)


cpvannier (commented):
Visually this is more legible, thanks.

However, this does not seem to fix the issue with some of the right click buttons like "Split tracklet" or "Add a point" that don't always work. I am not sure if this due to having to click in a specific area of the tracklet for the action to be triggered, or if this is something else.

BertrandRenault:
I found another issue related to this one, about dragging a tracklet (to expand/shrink it). I'm working on it so I put this PR on draft state.

cpvannier:
> > Visually this is more legible, thanks.
> > However, this does not seem to fix the issue with some of the right click buttons like "Split tracklet" or "Add a point" that don't always work. I am not sure if this due to having to click in a specific area of the tracklet for the action to be triggered, or if this is something else.
> 
> Does this have a dedicated issue ? If not could it be added so we can merge this PR and solve the issue mentioned here in a dedicated PR ?

Created a separate issue for it (#378) as we found the actual reason for this problem after discussing with Bertrand.

juliendenize:
- Sorry for the readiness I missclicked i put it back to draft
- Thanks for creating the other issue

BertrandRenault:
I fixed some issues with tracklet resizing (drag and drop the first/last key item of a tracklet), but there is still some problems on this.
As it is not the main isue of this PR, I open a new issue.

juliendenize (approved):

PR #377, opened 2024-12-03
bugfix unknown image type

BertrandRenault:
## Issue

image_to_base64 will crash with a "Unknown" image format.
It happens sometimes (I had it with one dataset. Images are .jpg but for it seems PIL doesn't correctly identify the format, although he can read it)

## Description

If format is "Unknown", use "JPEG" instead. As it is an internal save before base64 encoding, for thumbnail, this has no impact.


juliendenize (approved):

juliendenize:
It has no impact for thumbnails but this function is not dedicated to thumbnails, nonetheless i think it is an ok behavior.

PR #381, opened 2024-12-04
feat(back): Add back estimated dataset size

cpvannier:
## Description

Estimated dataset size was no longer calculated when loading datasets. Add it back.
Note: this estimated size no longer takes into account the media folder that can be stored outside the dataset.

### DIFF
--- a/pixano/datasets/dataset_info.py
+++ b/pixano/datasets/dataset_info.py
@@ -13,6 +13,7 @@ from typing import Literal, overload
 from pydantic import BaseModel, field_validator
 
 from pixano.features import Image
+from pixano.utils.python import estimate_folder_size
 
 
 class DatasetInfo(BaseModel):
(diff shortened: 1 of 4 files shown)


BertrandRenault:
Does this size have any interest without media ?
This size will mostly reflect lance table size, which could significantly change with or without embeddings.
In my opinion such metrics should be better in some computed stats
Number of bbox, mask, etc. should be more informative

cpvannier:
I just plugged the feature back, since the frontend was showing "Unknown" on all my datasets as it expected this size.

If this is considered as no longer relevant, it should be removed from both backend and frontend.

juliendenize:
This feature might indeed be misleading as this is not the size of the dataset but the size of the Lance tables as mentioned by Bertrand. Either we keep this and make it more explicit on what it is or we remove it.

If we keep it I worry it might slow down things because estimating the size suppose system calls which can be slow especially for a large number of artifacts by Lance.

That reminds me that we should take a look at compacting the datasets sometimes.

PR #383, opened 2024-12-09
change embeddings load

### LINKED ISSUE #382: SAM embedding from previous image kept after page change
## Bug description

After loading an image embedding, if we change current image, but stay on "interactive segmentation" tool, Pixano doesn't load current image embedding, so it segment with the previously displayed image embedding (last embedding loaded in fact)

## Steps to reproduce

1. Open a dataset with SAM embedding
2. Go to an image, activate "Interactive segmentation" tool. It load the model & embeddings
3. Go to another image, with left/right arrows
4. Click to segment: it segment with previously loaded embedding

image

Note that by changing tool, then back to "Interactive Segmentation", the model/embedding is asked and loaded again.

## Expected behavior

Coherent behaviour...

## Additional context

We should either force tool back to "Pan" (default) when changing image (dataset_item), or maybe auto load model and embedding (design change)

BertrandRenault:
## Issue

When changing dataset item, new embeddings is not loaded (previously loaded kept) until we open/close interactive segmentation tool.

Fixes #382 

## Description

Once a model + embedding has been loaded, model + embedding table is kept. New embedding are loaded on dataset item change

### DIFF
--- a/ui/apps/pixano/src/routes/+layout.svelte
+++ b/ui/apps/pixano/src/routes/+layout.svelte
@@ -31,10 +31,7 @@ License: CECILL-C
   let currentDatasetId: string;
   let currentDatasetItemsIds: string[];
 
-  async function handleGetModels() {
-    //TMP for lint
-    await new Promise((resolve) => setTimeout(resolve, 1));
-    modelsStore.set([]);
+  function handleGetModels() {
     api
       .getModels()
       .then((models) => modelsStore.set(models))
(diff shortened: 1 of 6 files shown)


ui/apps/pixano/src/routes/+layout.svelte
```diff
@@ -26,36 +26,24 @@ License: CECILL-C
...
-  }
-
-  onMount(async () => {
-    handleGetModels()
-      .then(() => console.log(`Found ${$modelsStore.length} model(s):`, $modelsStore))
-      .catch((err) => console.error("ERROR: Can't get model", err));
-    await handleGetDatasets();
+    api
+      .getDatasetsInfo()
+      .then((lodedDatasetInfos) => {
```
juliendenize, line 40: typo

juliendenize (changes requested):

juliendenize (approved):

PR #385, opened 2024-12-10
Dev/rename to textspan

benlabbe:
The category (ie concept_id) of the TextSpan is dropped , in favour of a referred Entity in the DatasetItem that could hold a category or have itself a Classification Annotation associated. This allows a clearer implementation of multi-modal-annotations (eg VQA with text-image grounding annotations). A future use-case text centred could require the re-introduction of the category in the TextSpan , but this is delayed.

## Issue

Fixes #


juliendenize (approved):
LGTM

PR #387, opened 2024-12-12
feat: add Message annotation as a registrerd base schema for the VQA usecase

benlabbe:


pixano/features/schemas/annotations/text_generation.py
```diff
@@ -0,0 +1,116 @@
...
+from .annotation import Annotation
+
+
+@_register_schema_internal
+class Message(Annotation):
+    """Textual exchange in a question/answer conversation
+    for image or text description and inforation extraction.
+
+    Attributes:
+        number: message number to associate diffrent ANSWER messages to a QUESTION.
```
BertrandRenault, line 25: diffrent -> different

pixano/features/schemas/annotations/text_generation.py
```diff
@@ -0,0 +1,116 @@
...
+        user: identify who is the author of the message (eg a human, a model, the ground truth, etc).
+        content: actual text of the message.
+        timestamp: creation date of the message.
```
BertrandRenault, line 28: there is already created_at & updated_at  timestamp in BaseSchema

BertrandRenault (approved):

pixano/features/schemas/annotations/text_generation.py
benlabbe: Good remark. @juliendenize  , do you think that the timestamp is still necessary in the schema ?

pixano/features/schemas/annotations/text_generation.py
benlabbe: fixed

pixano/features/schemas/annotations/text_generation.py
juliendenize: Yes to keep track of timestamp outside of Pixano (created is for Pixano) we could change later if we find there is no usecase for this

PR #405, opened 2025-01-13
feat(front): Better TimeTrack scaling

cpvannier:
## Description

Better scaling of markers and time indicators for longer videos with more granular options based on timeline density

image
image
image
image
image
image

Note: time indicators in "`{mm}`:`{ss}`" format for videos longer than a minute, and in "`{ss}`s" for shorter videos

### DIFF
--- a/ui/components/datasetItemWorkspace/src/components/VideoPlayer/TimeTrack.svelte
+++ b/ui/components/datasetItemWorkspace/src/components/VideoPlayer/TimeTrack.svelte
@@ -75,11 +75,25 @@ License: CECILL-C
   };
 
   const shouldDisplayTime = (ms: number, density: number) => {
-    if (ms % 10 !== 0) return false;
-    if (50 < density && ms % 50 === 0) return true;
-    if (25 < density && density < 50 && ms % 20 === 0) return true;
-    if (density < 25) return true;
-    return false;
+    if (density > 2000) {
+      return ms % 1200 === 0;
+    } else if (density > 1000) {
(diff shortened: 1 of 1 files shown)


ui/components/datasetItemWorkspace/src/components/VideoPlayer/TimeTrack.svelte
```diff
@@ -75,11 +75,25 @@ License: CECILL-C
...
+      return ms % 150 === 0;
+    } else if (density > 100) {
+      return ms % 100 === 0;
+    } else if (density > 50) {
+      return ms % 50 === 0;
+    } else if (density > 25) {
+      return ms % 20 === 0;
+    } else {
+      return ms % 10 === 0;
+    }
```
juliendenize, line 96: Couldn't we have an array of intervals aligned with an array of divider to make a more maintainable code instead of a long list of cases ? It should be equivalent in terms of perf and memory given the size of the arrays.

I wrote it in Python for convenience
```
densities = [2000, 1000, 500, 400, 250, 100, 50, 25]
dividers = [1200, 600, 300, 200, 150, 100, 50, 20]
for dens, div in zip(density, dividers):
   if density > dens:
     return (ms % div) == 0
return (ms % 10) == 0
```

cpvannier: Yes! Done

juliendenize (approved):

BertrandRenault (approved):

PR #429, opened 2025-01-23
Refresh thumbnails whenever needed

### LINKED ISSUE #428: When we validate tracks merge, thumbnails in ObjectInspector are not refreshed
## Bug description

When we validate tracks merge, thumbnails in ObjectInspector are not refreshed

BertrandRenault:
## Issue

Fixes #428 

## Description

thumbnails is now derived from entities, so any change made to this store is reflected in thumbnails

### DIFF
--- a/ui/components/datasetItemWorkspace/src/components/Inspector/ObjectCard.svelte
+++ b/ui/components/datasetItemWorkspace/src/components/Inspector/ObjectCard.svelte
@@ -277,8 +277,9 @@ License: CECILL-C
     !isEditing && selectedTool.set(panTool);
   };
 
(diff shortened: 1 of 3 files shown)


BertrandRenault:
commit name is wrong, should be something like PR title

BertrandRenault:
this was a bug, not directly related to thumbnails: list of entities in ObjectInspector where updated but not correctly refreshed when number of entities change (after create, delete or merge). In some case it would cause a mismatch between data displayed and actual items

--also, corrected wrongly set view_ref at entity creation in some cases

juliendenize (approved):

PR #436, opened 2025-01-31
Feat/multiple choices questions

antoningaladrim:
## Description

- Use consistent data in storybook
- Update vqa workspace layout
- Split open, single choice and multiple choice questions into different components
- Implement save logic


ui/components/vqaCanvas/src/lib/utils/groupMessagesByNumber.ts
```diff
@@ -0,0 +1,24 @@
...
+      return {
+        ...acc,
+        [groupKey]: [...(acc[groupKey] || []), item],
+      };
+    },
+    {} as { [key: string]: Message[] },
+  );
+
+  return Object.entries(groups)
+    .sort((a, b) => +a[0] - +b[0])
```
BertrandRenault, line 22: Any advantage of + signs here ?

antoningaladrim: Yes a[0] and b[0] are record keys therefore of type string

PR #445, opened 2025-02-06
feat: handle adding new answers to question

antoningaladrim:
### Issue

When no answer exists for a question, the inputs are not displayed

### Description

- Add function to create a new message corresponding to the answer of a user. The system works as follows :
  - If the current user already answered, the question is completed with his answer. If the user changes it, it updates the existing object in db
  - Otherwise, an empty answer is displayed by default and on the first submit, a new message for this user is created

- Update the Message model in the frontend to match with the definition of the backend

### DIFF
--- a/ui/components/datasetItemWorkspace/src/lib/utils/createNewAnswer.ts
+++ b/ui/components/datasetItemWorkspace/src/lib/utils/createNewAnswer.ts
@@ -1,0 +1,40 @@
+/*-------------------------------------
+Copyright: CEA-LIST/DIASI/SIALV/LVA
+Author : pixano@cea.fr
+License: CECILL-C
+-------------------------------------*/
+
+import { BaseSchema, Message, MessageTypeEnum, type MessageType } from "@pixano/core";
+import { nanoid } from "nanoid";
+
+export const createNewAnswer = ({
+  question,
+  content,
+  answers,
+  explanations,
+}: {
+  question: Message;
+  content: string;
+  answers?: string[];
+  explanations?: string[];
+}) => {
+  const now = new Date().toISOString();
+
+  const messageData: MessageType = {
+    number: question.data.number,
+    user: "user",
+    content,
+    timestamp: now,
+    type: MessageTypeEnum.ANSWER,
+    answers: answers ?? [],
+    explanations: explanations ?? [],
+  };
+
+  return new Message({
+    id: nanoid(10),
+    table_info: { name: "messages", group: "annotations", base_schema: BaseSchema.Message },
+    created_at: now,
+    updated_at: now,
+    data: { ...question.data, ...messageData },
+  });
+};
--- a/ui/components/datasetItemWorkspace/src/lib/utils/createNewAnswer.ts
+++ b/ui/components/datasetItemWorkspace/src/lib/utils/createNewAnswer.ts
@@ -9,18 +9,8 @@ import { nanoid } from "nanoid";
 import { sourcesStore } from "../../../../../apps/pixano/src/lib/stores/datasetStores";
 import { getPixanoSource } from "../api/objectsApi";
 
-export const createNewAnswer = ({
-  question,
-  content,
-  answers,
-  explanations,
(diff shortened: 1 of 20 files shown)


ui/components/datasetItemWorkspace/src/lib/utils/createNewAnswer.ts
```diff
@@ -0,0 +1,40 @@
...
+  question: Message;
+  content: string;
+  answers?: string[];
+  explanations?: string[];
+}) => {
+  const now = new Date().toISOString();
+
+  const messageData: MessageType = {
+    number: question.data.number,
+    user: "user",
```
antoningaladrim, line 25: TODO: how do you manage users @BertrandRenault ?

ui/components/datasetItemWorkspace/src/lib/utils/createNewAnswer.ts
```diff
@@ -0,0 +1,40 @@
...
+    answers: answers ?? [],
+    explanations: explanations ?? [],
+  };
+
+  return new Message({
+    id: nanoid(10),
+    table_info: { name: "messages", group: "annotations", base_schema: BaseSchema.Message },
+    created_at: now,
+    updated_at: now,
+    data: { ...question.data, ...messageData },
```
antoningaladrim, line 38: TODO: likewise, sourceRef should not be the same as the one of the question. How do you know the source ?

ui/components/datasetItemWorkspace/src/lib/utils/createNewAnswer.ts
BertrandRenault: In fact, Pixano app has no user management. So for now, when we need to put a source (which is the closer thing to a "user", but can also be a model or whatever) we use
(extract from defineCreatedObject.ts)
```
import { sourcesStore } from "../../../../../../apps/pixano/src/lib/stores/datasetStores";
import { getPixanoSource } from "./getPixanoSource";

 const pixSource = getPixanoSource(sourcesStore);
  const baseData = {
    item_ref: entity.data.item_ref,
    view_ref: viewRef,
    entity_ref: { name: entity.table_info.name, id: entity.id },
    source_ref: { name: pixSource.table_info.name, id: pixSource.id },  // <--- here
  };
```

but in this case, as "user" is a string... Either you put pixSource.id, either "user" (str) should be replaced by "source" (SourceRef)
@benlabbe do you think user here should be considered as a source ?

BertrandRenault: we use getPixanoSource, because we don't know who is the current user (for now) as there is no login management ...

ui/components/datasetItemWorkspace/src/lib/utils/createNewAnswer.ts
BertrandRenault: for source, the short answer is "we don't know it". So we use a generic source (aka "pixano") for everything that is created from Pixano app. - the getPixanoSource() mentionned [here](https://github.com/pixano/pixano/pull/445#discussion_r1944418499)

BertrandRenault: but ok, so "user" and source seems different... @benlabbe any insight on this "user" ?

BertrandRenault (approved):

ui/components/datasetItemWorkspace/src/lib/utils/createUpdatedAnswer.ts
```diff
@@ -6,19 +6,15 @@ License: CECILL-C
 
 import { Message } from "@pixano/core";
 
-export const createUpdatedMessage = ({
-  message,
-  newContent,
-  newChoices,
-  explanation,
+export const createUpdatedAnswer = ({
```
antoningaladrim, line 9: I realized this function was wrong because I do not change the updated_at property. I will fix it in another PR.

PR #455, opened 2025-02-10
Feat/update app header

antoningaladrim:
### Description

- Move tool bar and save button to header
- Update layout

image


ui/components/table/src/Table.svelte
```diff
@@ -134,9 +133,9 @@ License: CECILL-C
     </div>
   </div>
 </div>
+
 <div
-  class="h-[75vh] w-full overflow-y-auto overflow-x-auto
```
antoningaladrim, line 138: @BertrandRenault was this fixed size on purpose ?

BertrandRenault (commented):
I didn't investigate, but on this branch I got the following error on the console when opening any datasetItem of a video dataset:
image

ui/components/table/src/Table.svelte
BertrandRenault: oh you right. Then, I don't see any specific reason to have it fixed

PR #462, opened 2025-02-12
feat: Add COCO exporter

### LINKED ISSUE #420: Regression: Add back a base class to export annotations
## Feature description

With the new data model, there is no longer a base class to help export the annotations made with Pixano to a different format and be able to use them elsewhere. 

## Expected solution

Add back a base class that can help users code which annotations they want to export and in which output format, without having to interact directly with the Lance tables, in the same way that the DatasetBuilder helps with importing data.

This base class could be accompanied by a default exporter, for example to COCO format, which was available with the previous data model. 

The COCO format is widely used and its JSON structure is simple enough that it can be a good starting point to change the outputs to another data format for the users that need to.

cpvannier:
## Issue

Fixes #420

### Major changes
- Add `COCODatasetExporter` to export datasets to JSON files in COCO-format (with some additional fields)
  - Default COCO-format fields are followed strictly to try and ensure compatibility with COCO-format data loaders.
  - Some additional fields are included for reference:
    - In `info`: dataset `id` and `name`
    - In `annotations`: `pixano_entity_id`, `pixano_bbox_id`, `pixano_entity_id`. The annotation `id` is a combination of the view ID and entity ID to have unique IDs in the case of video datasets (where the entity is a track).
- Update `export_dataset_item` method of `DatasetExporter` to return `export_data` explicitly to prevent inconsistencies.
  - In the case of `COCODatasetExporter`, we have to merge information from Entities, BBoxes and CompressedRLEs, so we cannot use "call by sharing" to edit `export_data` in place with `append` and `extend`.
  - Also, for future exporters made by other Pixano developers or simply by users with specific needs, this should be easier to understand and to work with.
### Minor changes
- Add `mask` property in `CompressedRLE` for COCO-format export
- Update `labels` utils for exporting categories to COCO-format
- Split JSON and JSONL exporters into two separate files
- Update exporters tests


pixano/datasets/exporters/coco_dataset_exporter.py
```diff
@@ -0,0 +1,225 @@
...
+            "area": None,
+            "bbox": None,
+            "confidence": None,
+            "iscrowd": 0,
+            "pixano_entity_id": ann.id if isinstance(ann, Entity) else ann.entity_ref.id,
+        }
+    )
+    # Add the specific elements from the annotation into the COCO format
+    if isinstance(ann, Entity):
+        category_name = str(ann.category).strip().lower()
```
BertrandRenault, line 211: Probably I missed something, but how can you know "category" is an existing field in Entity ?
The base Entity doesn't have a "category" field (only Track have a default "name" field)

cpvannier: There is a condition on line 114 to enter this function:

```python
if isinstance(schema, Entity) and hasattr(schema, "category"):
```

pixano/app/serve.py
```diff
@@ -200,14 +200,12 @@ def get_env(self) -> str:
...
-    help=("S3 endpoint URL, use 'AWS' if not provided. " "Used if library_dir or media_dir is an S3 path"),
+    help=("S3 endpoint URL, use 'AWS' if not provided. Used if library_dir or media_dir is an S3 path"),
 )
 @click.option(
     "--aws_region",
     type=str,
-    help=(
-        "S3 region name, not always required for private storages." "Used if library_dir or media_dir is an S3 path"
-    ),
+    help=("S3 region name, not always required for private storages.Used if library_dir or media_dir is an S3 path"),
```
juliendenize, line 208: There is a space lacking

juliendenize (changes requested):
Can you merge main ? apparently there is a conflict to resolve

juliendenize (approved):

PR #463, opened 2025-02-12
Feat/update app header

antoningaladrim:
## Description

- Update the header of Pixano app
  - Move toolbar and save button to header
  - Change style
- Refactor datasetExplorer creating DatasetBrowserForm and DatasetPagination components


ui/apps/pixano/src/components/layout/DatasetItemHeader.svelte
```diff
@@ -0,0 +1,57 @@
...
+      <IconButton
+        on:click={() => goToNeighborItem("next")}
+        tooltipContent="Next item (shift + right arrow)"
+      >
+        <ArrowRight />
+      </IconButton>
+    </div>
+
+    <Toolbar />
+    <!-- <Toolbar isVideo={selectedItem.ui.type === WorkspaceType.VIDEO} /> -->
```
cpvannier, line 53: `isVideo` is needed to display the FusionTool in Toolbar

cpvannier (changes requested):

cpvannier:
Also, a suggestion.

Now that the tools are stacked horizontally, the misalignment because of the `MagicIcon` on the bottom left of the smart segmentation tools is even more visible:
image

My proposition is to remove this `MagicIcon` altogether, and replace the `BrushIcon` with the `Wand2Icon` to keep the "magic" element:

### Before

image

### After

image

@BertrandRenault, @juliendenize, thoughts?

If this needs some discussion, I'm okay to move this to another issue/PR and merge this one as is. :+1:

ui/apps/pixano/src/components/layout/DatasetItemHeader.svelte
```diff
@@ -26,6 +29,11 @@ License: CECILL-C
     }
     return event.key;
   };
+
+  const isVideoDatasetItem = async () => {
+    const datasetItem = await api.getDatasetItem($currentDatasetStore.id, currentItemId);
+    return datasetItem.ui.type === WorkspaceType.VIDEO;
```
antoningaladrim, line 35: @cpvannier when I log datasetItem.ui.type I always get "undefined" (as a string)

cpvannier: Same here. It's because `datasetItem.ui.type` is not set in `api.getDatasetItem`, but in `handleSelectItem` in [[itemId]/+page.svelte](https://github.com/pixano/pixano/blob/908c71040120ccc5b8718a1ed65d9a66652225c8/ui/apps/pixano/src/routes/%5Bdataset%5D/dataset/%5BitemId%5D/%2Bpage.svelte#L55).


You could use:
```svelte
    <Toolbar isVideo={$currentDatasetStore.workspace == WorkspaceType.VIDEO} />
```

But looking at `handleSelectItem`, sometimes `dataset.workspace` can be undefined, and then `datasetItem.ui.type` is inferred from the item content.

cpvannier (changes requested):
The FusionTool appears!

- FusionTool should be aligned with the other tools (`mt-4` in FusionTool.svelte:122)
image

- FusionTool options should now be displayed horizontally
image

cpvannier (approved):

cpvannier (approved):

ui/components/datasetItemWorkspace/src/components/Toolbar/KeyPointsSelectionTool.svelte
```diff
@@ -56,10 +56,12 @@ License: CECILL-C
...
-          <div
-            class="w-1 h-1 bg-primary rounded-full absolute"
-            style={`top: ${vertex.y * 100}%; left: ${vertex.x * 100}%; background: ${defineDotColor(vertex, template.template_id === $selectedKeypointsTemplate)}`}
-          />
+          <div>
+            <div
+              class="w-1 h-1 bg-primary rounded-full absolute"
+              style={`top: ${vertex.y * 100}%; left: ${vertex.x * 100}%; background: ${defineDotColor(vertex, template.template_id === $selectedKeypointsTemplate)}`}
+            />
+          </div>
```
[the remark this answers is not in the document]
cpvannier: This change is on me! There definitely should be better way to do this, but basically since the divs are positioned as `absolute`, the parent that they use as reference without this extra `div` probably isn't setup correctly.

I tried looking at https://tailwindcss.com/docs/position#absolutely-positioning-elements but couldn't find what was wrong.

### With the extra `div`
image

### Without 
image

PR #467, opened 2025-02-18
Feat/add question form

antoningaladrim:
## Description

- Add a form to add a question to a vqa item
- Add a button to the vqaCanvas to open the form

## TODO

- Fetch available models from backend
- Generate genuine question with pixano inference


ui/components/datasetItemWorkspace/src/components/DatasetItemViewer/VqaViewer.svelte
```diff
@@ -150,12 +152,30 @@ License: CECILL-C
       updateMessageContent(event.detail);
     }
   };
+
+  const handleStoreQuestion = (event: CustomEvent<StoreQuestionEvent>) => {
+    const datasetItemEntities = selectedItem.entities;
+    const conversationEntities = Object.values(datasetItemEntities)
```
antoningaladrim, line 158: Is there a cleaner way to achieve this ? I did not find any reference of the current entity

ui/components/datasetItemWorkspace/src/components/DatasetItemViewer/VqaViewer.svelte
```diff
@@ -150,12 +152,30 @@ License: CECILL-C
...
+      .flat()
+      .filter((entity) => entity.is_type(BaseSchema.Conversation));
+
+    if (conversationEntities.length === 0) {
```
antoningaladrim, line 162: This assumes there is always a conversation entity which I guess should be the case, do we implement the creation of one if it is not the case ?

ui/components/datasetItemWorkspace/src/components/DatasetItemViewer/VqaViewer.svelte
BertrandRenault: at this point, $entities (the store) is already fill, so you can filter for conversations

`const conversationEntities = $entities.filter((e)=>e.is_conversation())`

alternatively, you can create a derived store (pro (&con): it is updated whenever $entities change, and it is persistent if you need $conversations often)

```
import { derived } from "svelte/store";
const conversations = derived(entities, ($entities)=>$entities.filter((e)=>e.is_conversation()));
```

ui/components/datasetItemWorkspace/src/components/DatasetItemViewer/VqaViewer.svelte
BertrandRenault: Not too sure about this.
In my opinion we should implement the creation (as for entities when drawing a new annotation, cf saveShapeForm.svelte and defineCreatedEntity.ts). But maybe we could discuss this

BertrandRenault (dismissed):

BertrandRenault (approved):

BertrandRenault (approved):

PR #470, opened 2025-02-20
inference_metadata field front fixes + object total count fix

BertrandRenault:
## Issue

- total count of object wrongly counts Conversations
- inference_metadata is wrongly asked to user at creation time
- inference_metadata is not set at creation time, which lead to a zod error

## Description

- remove inference_metadata field from user's features,
- set default inference_metadata value at creation
- do not count conversations in Objects total count.


BertrandRenault:
Note: in ObjectCard.svelte, there is only one real change: removal of :
`{#if entity.table_info.name !== "conversations"}` and its final `{/if}`
All other displayed changes are because of indent style

juliendenize (approved):

PR #477, opened 2025-02-24
Fix/dataset explorer

antoningaladrim:
## Description

In the datasetExplorer page, the table does not update on pagination change. Fixed by dynamically setting the content of the table 

@cpvannier, @BertrandRenault I definitely do not think it is a clean fix, feel free to suggest anything better

### DIFF
--- a/ui/apps/pixano/src/components/layout/DatasetHeader.svelte
+++ b/ui/apps/pixano/src/components/layout/DatasetHeader.svelte
@@ -6,7 +6,7 @@ License: CECILL-C
 
 <script lang="ts">
   // Imports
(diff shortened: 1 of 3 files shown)


BertrandRenault:
In fact it's probably a better way than before, which reloaded the component.

I made some minor cosmetic changes:

- use pixano logo as an icon to go back to Home page
image

- put back the back arrow for "back to dataset"
image

- and also, display dataset name even in dataset item page, not only in dataset explorer

BertrandRenault (approved):

BertrandRenault:
TODO:  (for compliance with design) Home icon back in DatasetItemHeader, remove dataset name. Keep "back to dataset"

BertrandRenault (approved):

PR #481, opened 2025-02-25
add a "cancel" button to SelectModal

BertrandRenault:
## Issue

In SelectModal (used when we select a segmentation model and embeddings), when no choice is available, or if we want to abort, only a "OK" button. In some case it cause an infinite loading wheel (though nothing is loading)

## Description

- Add a Cancel button
- disable OK button is nothing is selected

### DIFF
--- a/ui/components/core/src/components/modals/SelectModal.svelte
+++ b/ui/components/core/src/components/modals/SelectModal.svelte
@@ -51,7 +51,7 @@ License: CECILL-C
 
       <button
         type="button"
-        disabled={!selected || selected === ""}
+        disabled={!selected}
         class="rounded border border-transparent text-slate-50 mt-3 mx-1 py-1 px-3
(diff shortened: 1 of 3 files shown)


ui/components/core/src/components/modals/SelectModal.svelte
```diff
@@ -47,12 +51,21 @@ License: CECILL-C
 
       <button
         type="button"
+        disabled={!selected || selected === ""}
```
antoningaladrim, line 54: !("") === true, I think the second assertion is redundant

antoningaladrim (dismissed):

ui/components/core/src/components/modals/SelectModal.svelte
BertrandRenault: You right, my bad !

PR #482, opened 2025-02-26
Add MultimediaEntity for the MultimediaEntityLinking usecase

benlabbe:
## Description

Add MultimediaEntity for the MultimediaEntityLinking usecase


juliendenize (changes requested):
I have two requests:
- Create a class named `NamedEntity` and make `Track` and `MultiMediaEntity` inherit from it. It avoid the duplication of having name attribute while giving the opportunity to users to directly use a named entity outside of the scope of tracks or multi media.
- Test the classes and their associated functions (is_, create_) with the helpers provided in the test package.

benlabbe (commented):
The MultimediaEntity class should be renamed into MultiModalEntity

benlabbe (commented):
Seems good to me

BertrandRenault (approved):

PR #489, opened 2025-03-03
dataset thumbnail: get absolute path

BertrandRenault:
## Issue

thumbnails displayed in datasets explorer page do not resolve relative path, and so doesn't display thumbnail if library dir is given as a relative path

## Description

resolve() path to absolute

### DIFF
--- a/ui/apps/pixano/src/components/dataset/DatasetPreviewCard.svelte
+++ b/ui/apps/pixano/src/components/dataset/DatasetPreviewCard.svelte
@@ -58,7 +58,7 @@ License: CECILL-C
   <div class="m-4 bg-slate-50">
     <img
       src={dataset.preview ? dataset.preview : pixanoLogo}
-      alt="{dataset.name} {dataset.preview ?? 'pixanoLogo'} thumbnail"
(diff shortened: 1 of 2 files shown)


ui/apps/pixano/src/components/dataset/DatasetPreviewCard.svelte
```diff
@@ -57,7 +57,7 @@ License: CECILL-C
   <!-- Dataset Thumbnail -->
   <div class="m-4 bg-slate-50">
     <img
-      src={dataset.preview ?? pixanoLogo}
+      src={dataset.preview ? dataset.preview : pixanoLogo}
```
antoningaladrim, line 60: This is the same thing as previously isnt it ?

antoningaladrim (approved):

ui/apps/pixano/src/components/dataset/DatasetPreviewCard.svelte
BertrandRenault: in fact no. "??" only check for null and undefined, but ? also resolve to false with an empty string (which is the case here when no thumbnail)

PR #490, opened 2025-03-04
fix regex vite.config

BertrandRenault:
## Issue

regex used in vite.config.ts do not properly handle routes prefix "models" and "app_models"

## Description

- fixed regex
- remove some dev logs

### DIFF
--- a/ui/apps/pixano/src/routes/[dataset]/dataset/[itemId]/+page.svelte
+++ b/ui/apps/pixano/src/routes/[dataset]/dataset/[itemId]/+page.svelte
@@ -65,7 +65,7 @@ License: CECILL-C
   type backFVS = Record<string, Record<string, backFV[]>>;
 
   const mapBackFeaturesValues2FrontFeaturesValues = (feature_values: backFVS) => {
-    console.log("--- back feature_values", feature_values);
+    //console.log("--- back feature_values", feature_values);
     /*
     --- current back format for features values (most probably to be changed)
     {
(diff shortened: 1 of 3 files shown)


ui/apps/pixano/src/routes/[dataset]/dataset/[itemId]/+page.svelte
```diff
@@ -65,7 +65,7 @@ License: CECILL-C
   type backFVS = Record<string, Record<string, backFV[]>>;
 
   const mapBackFeaturesValues2FrontFeaturesValues = (feature_values: backFVS) => {
-    console.log("--- back feature_values", feature_values);
+    //console.log("--- back feature_values", feature_values);
```
Comment (antoningaladrim) [inline ui/apps/pixano/src/routes/[dataset]/dataset/[itemId]/+page.svelte:68]: we could remove the line instead if we do not use it

antoningaladrim (approved):

ui/apps/pixano/src/routes/[dataset]/dataset/[itemId]/+page.svelte
Comment (BertrandRenault) [reply ui/apps/pixano/src/routes/[dataset]/dataset/[itemId]/+page.svelte:68]: this whole function is a mapping of a json format that should be revised to an old back format (before datamodel changes) which does fulfill common needs, but is not robust to edge cases (non unique features names on different tables)
This need rework (https://github.com/pixano/pixano/issues/491) so I think it's ok if it looks "temporary". This is a WIP piece of code, only because theses "common needs" are asked for a too long time.

PR #499, opened 2025-03-12
modal prompt & temperature

BertrandRenault:
## Description

- Allow user to enter prompts used for question/answer generation
(pre-prompt for answer, the prompt is the question asked)

- Allow user to set temperature

- (back) Change waw of gathering images and bbox. Previously it was based on a image tag regex, now we send all images of the current dataset item, regardless of any image tag. Also, we now send all bbox coordinates as well as it's name (if exist)


benlabbe (approved):
OK for me

PR #511, opened 2025-03-25
textspan: compute background luminace to choose text color

BertrandRenault:
## Issue

With some colors, text becomes hard to read
Image

## Description

Compute luminance of background color to choose black or white font color
image


benlabbe (approved):
Nice upgrade

PR #512, opened 2025-03-26
fix(front): Fix issues with fusion tool

cpvannier:
## Description

- Disable highlighting object on image/video canvas for all editing tools (all tools except the pan tool), not just for the fusion tool
- TODO: Fix issue where objects stay highlighted when going from fusion tool to pan tool, despite highlighting being reset in cleanFusion


ui/components/datasetItemWorkspace/src/components/Toolbar.svelte
```diff
@@ -57,10 +58,7 @@ License: CECILL-C
 
   const selectTool = (tool: SelectionTool) => {
     if (tool !== $selectedTool) {
-      if (previousSelectedTool?.type === ToolType.Fusion) {
-        //abort fusion
-        cleanFusion();
-      }
+      cleanFusion();
```
BertrandRenault, line 61: as now cleanFusion is a generic unhighlight / unedit all function, we should rename to unHighlightAll, or cleanHighlights maybe ?

cpvannier: Yes, it would be more clear

cpvannier:
Just tested the latest changes, the issue is now fixed, thanks!
Ok to merge!

BertrandRenault (approved):

PR #520, opened 2025-03-31
Set none limit to table size

BertrandRenault:
## Issue

Dataset.get_data() require a limit (if ids and item_ids not set)
But sometimes we don't know a suitable limit, (ex: source - only a few rows expected).
Also, lancedb now warn when scanning out of bound:
[2025-03-31T09:57:05Z WARN  lance::io::exec::scan] Ran out of fragments before we were done scanning for range: 0..1000

Anyway, it's not convenient to have to set a limit when you don't know yet how much rows the table contains.

## Description

Allows to have unset limit: in this case, limit is set to the table size (table.count_rows())


benlabbe (approved):
OK for me

PR #521, opened 2025-03-31
filter unique mentions

BertrandRenault:
## Issue

ObjectCard shows one "mention" line per TextSpan, although mention could be the same.

## Description

Filter display by unique mention


benlabbe (approved):
OK

PR #524, opened 2025-04-01
allow schema overwrite

BertrandRenault:
## Issue

If a schema already exist in schema.json, raise an error. 
It is really annoying for SAM embeddings, when something went wrong while computing them (eg. Out of memory error)

## Description

Allow shema overwrite

### DIFF
--- a/pixano/datasets/dataset.py
+++ b/pixano/datasets/dataset.py
@@ -237,7 +237,7 @@ class Dataset:
             schema: Table schema.
             relation_item: Relation with the `'item'` table (table to item).
             data: Table data.
-            mode: Table mode ('create', 'overwrite'd).
+            mode: Table mode ('create', 'overwrite').
             exist_ok: If True, do not raise an error if the table already exists.
             on_bad_vectors: Raise an error, drop or fill bad vectors ("error", "drop", "fill").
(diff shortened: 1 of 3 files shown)


benlabbe:
I am not sure that this behavior is justified.
Could we imagine that the error is raised only in cases were we know that the targeted schema-table is expected to be empty (for brand-new new dataset , or when the 'overwrite' option of the dataset builder is not toggled) ?

cpvannier (commented):
`create_table()` already has a `mode` argument that can be set to `"overwrite"` and an `exist_ok` argument that can be set to `True`.

Why not pass one of those argument to `add_schema()`, so that an existing schema triggers a `ValueError` by default, but can be overwritten if `mode == "overwrite"` / `exist_ok == True`.

cpvannier (commented):
I didn't know the different between `mode="overwrite"` and `exist_ok=True` so I thought they were redundant, but after looking at the LanceDB documentation, I understand that:
- `"overwrite"` means it's okay if the table exists, and it will be replaced by a new one
- `exist_ok` means it's okay if the table exists, and it will be used as is

I guess that we should use `exist_ok or mode == "overwrite"` instead of `and` for `overwrite_schema` then, as there might be different use cases for each of them.

PR #528, opened 2025-04-02
remove same type limitation

BertrandRenault:
## Issue

Relink do not allow an entity with child of same type. It was a security check that is not relevant, and this behaviour could be needed.

## Description

Remove same type check.


BertrandRenault:
close (done in #531)

PR #531, opened 2025-04-03
allows SAM1 in video (not SAM2 yet)

### LINKED ISSUE #214: Mask annotation in video frame

BertrandRenault:
## Issue

Currently we can't use interactive segmentation in video

Fixes #214 

## Description

Allows interactive segmentation in video.
As we can't easily interpolate mask, this need another PR to be OK (to create 1 frame tracklet - link soon)

### DIFF
--- a/ui/components/datasetItemWorkspace/src/components/DatasetItemViewer/DatasetItemViewer.svelte
+++ b/ui/components/datasetItemWorkspace/src/components/DatasetItemViewer/DatasetItemViewer.svelte
@@ -67,7 +67,7 @@ License: CECILL-C
       <Loader2Icon class="animate-spin text-white" />
     </div>
   {:else if selectedItem.ui.type === WorkspaceType.VIDEO}
-    <VideoViewer {selectedItem} {resize} bind:currentAnn />
+    <VideoViewer {selectedItem} {embeddings} {resize} bind:currentAnn />
   {:else if selectedItem.ui.type === WorkspaceType.IMAGE_VQA}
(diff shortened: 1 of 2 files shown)


ui/components/datasetItemWorkspace/src/components/VideoPlayer/ObjectTracklet.svelte
```diff
@@ -68,19 +77,80 @@ License: CECILL-C
...
+  });
+
   $: color = $colorScale[1](trackId);
 
   $: tracklet_annotations_frame_indexes = tracklet.ui.childs.map((ann) => ann.ui.frame_index!);
 
+  $: canSplitOrAddKeyFrame =
+    $currentFrameIndex > tracklet.data.start_timestep &&
+    $currentFrameIndex < tracklet.data.end_timestep &&
+    !tracklet.ui.childs.some((ann) => ann.ui.frame_index === $currentFrameIndex);
```
cpvannier, line 113: This condition makes sense for adding a key frame, however I think you should be able to split a tracklet even if there are some annotations on the current frame.

One tracklet can end on the frame `n` with the next one beginning on the frame `n+1`, so it is possible to split without an empty frame between the two.

If the confusion is about if the current frame `n` should belong to the left tracklet or the right tracklet, maybe we could add options like "Split before current frame" and "Split after current frame".

cpvannier (commented):
I think the condition for splitting a tracklet should be fixed (see the comment in ObjectTracklet.svelte), but other than that segmentation in video datasets is working for me. :+1:

cpvannier (approved):

PR #533, opened 2025-04-09
chore(front): Remove capitalization of names of object features

cpvannier:
## Description

To prevent weird capitalization like "Category Id" instead of either "ID" or "id", don't capitalize the names of object features and leave the text in lowercase, as is already done in the DatasetExplorer Table.


BertrandRenault (approved):
Agreed, it's quite nonsense to modify user choosen display

PR #537, opened 2025-04-09
TextSpan enhancements

BertrandRenault:
## Issue

With textSpan of composed words, the HTML span only take the first word (+ 1 char !)

Need to be able to select a TextSpan, and see it is selected.

## Description

- Fixed textSpan color highlight (now, it take each word (internal span) separatly, but glue them if they follow
  (end[N] - start[N+1] <= 2  --> this allow 1 or two chars between two words (aka a space, or punctuation + space))

- TextSpan can now be selected in TextView (by a click) -> it highlight them (with a glow effect) ans also select the corresponding entity

- also fix line feeds (\n -> \<br>)

### DIFF
--- a/ui/components/textCanvas/src/TextSpanArea.svelte
+++ b/ui/components/textCanvas/src/TextSpanArea.svelte
@@ -46,16 +46,18 @@ License: CECILL-C
   };
 </script>
 
-<div class="bg-white p-2 flex flex-col gap-2 h-full overflow-y-auto">
+<div class="bg-white p-2 flex flex-col gap-2 h-full">
   <button class="bg-primary text-white p-2 rounded-md w-fit" on:click={onTagText} id="tagButton">
     Tag Selected Text
   </button>
-  {#each textViews as textView}
-    <SpannableTextView
-      {textView}
-      {colorScale}
-      textSpans={spansByViewId[textView.id]}
-      bind:textSpanAttributes
-    />
-  {/each}
+  <div class="overflow-y-auto">
+    {#each textViews as textView}
+      <SpannableTextView
+        {textView}
+        {colorScale}
+        textSpans={spansByViewId[textView.id]}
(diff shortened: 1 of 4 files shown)


benlabbe:
In the case of comosed words test-spans, the correct representation is to store the start offset and end offset of each word, so that it is theoricaly possible to express a composed word with a hole/elision. Exemple, selecting only "grand château de sable" as the concept of interest in the following sentence :
" le plus grand et le plus beau château de sable"

I admit that this kind of elision is a rare use case, and that the UI to select/exclude/merge multiple words into a single Text-span would be difficult to implement.
But until we have a  pressure to implementing such a UI feature, it is still possible to comply with the storing format of text spans for single word and sequential composed words.

The storing format is a couple of related lists of offsets. To store the spans of the sequential composed word "chateau de sable" in the next  sentence "construis un château de sable avec moi", the couple of related offsets could be :

span_starts : list[int] = [ 14, 22, 24 ]
span_ends : list[int] = [ 20, 23, 29 ]

Or 

span_starts : list[int] = [ 14 ]
span_ends : list[int] = [  29 ] 

Both are correct. 

And the displayed highlighted span could be merged into a single rectangle, from offset 14 to 29. 

NB, the offsets might be 0 indexed, so consider that  the values in this exemple are decreased by 1.

BertrandRenault:
> span_starts : list[int] = [ 14, 22, 24 ] span_ends : list[int] = [ 20, 23, 29 ]

Before this fix, the resulting span was 14 -> 21 (20+1)  (= only the first word + 1char)
With this fix, the resulting span is 14 -> 29

Which is correct for "chateau de sable", but not for "grand château de sable" (-> "grand et le plus beau château de sable")
I will try to highlight each word independantly so

benlabbe (approved):
Great commit

PR #539, opened 2025-04-11
Allow filtering of datasetItems on browser page

BertrandRenault:
## Description

Allow to filter dataset items in browser page.

A selector allows to choose a column amongst columns (restricted to string/int/float types)
An input field allows to enter the value to seek in this column (check for equality)

Typing enter in this input launch the filter

A button allows to clear the filter

A "special" free mode allows to enter a "where" clause directly (for more specific / complex / composed filter)

Also, the filtered selection is kept when navigating in dataset Item page.


benlabbe (approved):
OK for me. But could not test myself

PR #540, opened 2025-04-14
do not select tracklet on right click when fusion

### LINKED ISSUE #534: Right-clicking a tracklet with FusionTool selected resets object association
## Bug description

When associating objects with the FusionTool, if you right-click a tracklet in the timeline, e.g. to split it or delete it, it resets the list of objects selected for association.

## Steps to reproduce

1. Open a video dataset
2. Click the fusion tool
3. Click a few objects to select them for association
4. Right-click a tracklet in the timeline

## Expected behavior

You should be able to right-click a tracklet to access options like splitting or deleting without losing your list of objects selected for association.

BertrandRenault:
## Issue

Right clicking on a tracklet (in VideoInspector) while in fuison mode lose current selection.

Fixes #534 

## Description

Prevent selection of right-clicked tracklet when in fusion mode.

Manage delete in fusion mode (update "to fuse" and "forbidden" lists)

Also, removed ability to relink & glue while in fusion mode, as it could create confusion with the current "to_fuse" selection.

@cpvannier In fact I should also probably remove split, as it also may create buggy cases. Another option is to limit split & glue to tracks outside of the ones selected for fusion, but it may change the forbidden list too, so need a specific manangement. I think it's better to just disallow them. (For relink, it's too confusing to relink while fusing -- which is the same thing in fact).

@cpvannier Another note: the relink feature is now more permissive than the fusion one. (it allow overlap of interpolated tracklet, if no keyframes overlap) We should discuss this maybe ?

### DIFF
--- a/ui/components/datasetItemWorkspace/src/components/Inspector/ChildCard.svelte
+++ b/ui/components/datasetItemWorkspace/src/components/Inspector/ChildCard.svelte
@@ -147,16 +147,16 @@ License: CECILL-C
       >
         <Pencil class="h-4" />
       </IconButton>
+      <IconButton
+        tooltipContent="Relink object"
+        selected={showRelink}
+        on:click={() => {
+          showRelink = !showRelink;
+        }}
+      >
+        <Link class="h-4" />
+      </IconButton>
     {/if}
-    <IconButton
-      tooltipContent="Relink object"
-      selected={showRelink}
-      on:click={() => {
-        showRelink = !showRelink;
-      }}
-    >
-      <Link class="h-4" />
-    </IconButton>
     <IconButton
       tooltipContent="Delete object"
       redconfirm
--- a/ui/components/datasetItemWorkspace/src/components/VideoPlayer/ObjectTrack.svelte
+++ b/ui/components/datasetItemWorkspace/src/components/VideoPlayer/ObjectTrack.svelte
@@ -98,7 +98,7 @@ License: CECILL-C
   };
 
   const onContextMenu = (tracklet: Tracklet | null = null) => {
-    if (tracklet) {
+    if (tracklet && $selectedTool.type !== ToolType.Fusion) {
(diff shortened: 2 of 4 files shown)


cpvannier:
> @cpvannier In fact I should also probably remove split, as it also may create buggy cases. Another option is to limit split & glue to tracks outside of the ones selected for fusion, but it may change the forbidden list too, so need a specific manangement. I think it's better to just disallow them. (For relink, it's too confusing to relink while fusing -- which is the same thing in fact).

I think it is better to disallow those options completely then, to prevent confusion and possible bugs.

> @cpvannier Another note: the relink feature is now more permissive than the fusion one. (it allow overlap of interpolated tracklet, if no keyframes overlap) We should discuss this maybe ?

It should be the same criteria for both in my opinion. I guess there's no reason not to make the fusion feature as permissive as the relink feature?

BertrandRenault:
> It should be the same criteria for both in my opinion. I guess there's no reason not to make the fusion feature as permissive as the relink feature?
image
In this image, relink allow to fuse theses 2 tracks (the blue one only has 2 keyframes, at start & end)
Fusion doesn't allow it (no overlaping at all). Somehow I agree we should have common behaviour, but from the fusion tool it seems strange to have more than one selected item per view (at frame X)

cpvannier:
> but from the fusion tool it seems strange to have more than one selected item per view (at frame X)

I agree that in general we wouldn't want to allow that, but the scenarios where it is possible (like the one you're showing) are some edge cases where two tracklets overlap even though they have no keyframes in common.

I think we can give the option to the user in those edge cases, while in most cases it won't be possible.

cpvannier (approved):
Opened #552 (Remove the ability to split tracklets while in fusion mode) and #553 (Use the same criteria for relink and fusion) for the remaining issues, so that this PR can still be merged in the meantime

PR #541, opened 2025-04-16
add index of count in header

BertrandRenault:
## Issue

When annotating (from DatasetItem view), it's convenient to know where we are in the dataset items.

## Description

Add "{index} of {num_items}"  in the header.
It take account of the current filter if any


benlabbe (approved):
OK for me

cpvannier (approved):

PR #547, opened 2025-05-13
fix position and avoid linefeed replace

### LINKED ISSUE #546: TextSpan highlights are shifted below the first line
## Bug description

In the VQA workspace . The text display has two defect behaviours .

1. Once text is selected to add a TextSpan, we observe a mis-alignment of the TextSpan offset between what was selected and what is retrieved and highkighted.
2. The pattern replacement for a user-friendly text display is not executed in the UI at the first display of the DatasetItem.

## Steps to reproduce

1. Open a VQA dataset and a DatasetItem which text contains multiple lines (multiple `\n`). Some TextSpan may be already present, from the dataset creation. These TextSpan are correctly highlighted.
2. Select a new word (or compound-word) in the text area on the left near the end of the text, and tag it as a new TextSpan. Give it the name you want.
3. Observe that the highlighted characters of this TextSpan are shifted. The shift is larger if the selected words are fare below the first text line.
4. The TextSpan annotation in the database does not contain the correct word selected in the UI.

## Expected behavior

The highlighted colored box should surround exactly the characters selected when creating the new TextSpan.
The TextSpan annotation in the database should contain exactly the selected word(s) .

## Comment

1. The defect is certainly associated to the on-the-fly replacement of the ASCI  `\n`  by  HTML `<br/>` for a user-friendly text display.
These two patterns don't have same length , so the positions (start-end)  found in the TextSpan annotations should be interpreted only in the original text (before the replacement).

BertrandRenault:
## Issue

Incorrect TextSpan position since a previous change

Fixes #546 

## Description

- remove incorrect "-1" for textspan length
- remove the replaceAll("\n",  "\<br>") that changed positionning, instead using "whitespace-pre-wrap" div class

### DIFF
--- a/ui/components/textCanvas/src/components/SpannableTextView.svelte
+++ b/ui/components/textCanvas/src/components/SpannableTextView.svelte
@@ -47,7 +47,7 @@ License: CECILL-C
 <div
   on:mouseup={mouseupListener}
   on:mousedown={clickHandler}
-  class="border rounded-lg p-2"
+  class="border rounded-lg p-2 whitespace-pre-wrap"
   role="textbox"
   tabindex="0"
 >
--- a/ui/components/textCanvas/src/components/SpannableTextView.svelte
+++ b/ui/components/textCanvas/src/components/SpannableTextView.svelte
@@ -20,7 +20,17 @@ License: CECILL-C
   export let textSpanAttributes: TextSpanType | null = null;
   export let textView: TextView;
 
-  $: richEditorContent = textSpansToHtml({ text: textView.data.content, textSpans, colorScale });
+  function escapeHtmlTagsOnly(text: string): string {
+    return text.replace(/<\/?[a-zA-Z][^<>]*?>/g, (match) => {
+      return match.replace("<", "-").replace(">", "-");
+    });
+  }
+
+  $: richEditorContent = textSpansToHtml({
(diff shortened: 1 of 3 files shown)


benlabbe (commented):
I tested this branch, but the issue is not fixed completely from my point of view.

I still observe a bad positioning of the highlighting rectangle on the span , especially if the new span is created at the end of a long document.
The bug is not observed if the new span is created within the first characters of the document.

BertrandRenault:
> I still observe a bad positioning of the highlighting rectangle on the span , especially if the new span is created at the end of a long document. The bug is not observed if the new span is created within the first characters of the document.
 
This text contains an unwanted tag "\<i>".
We have to display text as HTML to add colored textspan. But if the original text contains some "tags", it's obviously an issue.
I add a replacer for this kind of "fake" HTML tag. I'm not happy with it but it seems escaper '\' is not a good solution here. So we replace "<...>" by "-...-" (in displayed text only)

benlabbe:
Nice improvement. Congratulations.

I still can see the bug when we have mentions (TextSpan) that span a linefeed. The future TextSpan annotated after this one will be ill placed

benlabbe (approved):
OK for now.

The problem still exists when the mention spans on multiple lines (linefeed included inside the mention) . This could be fixed later

PR #549, opened 2025-05-16
fix builder when id or split in metadata.jsonl

BertrandRenault:
## Issue

If "id" or "split" is in metadata.jsonl (or any genrated dict used for folder builders), _create_items raise a "TypeError: ... got multiple values for keyword argument ..."

## Description

change _create_items and calls to this function so that "id" and "split" are not parameters but already in the item_metadata dict


benlabbe (approved):
OK for me

PR #550, opened 2025-05-16
feature update: missing MultiModalEntity in allowed Entities

BertrandRenault:
## Issue

In a EntityLinking workspace, features edition doesn't work

## Description

In EntityLinking, entities are MultiModalEntity, and where not correctly handled by previous filter (only on Track and Entity)
Added MultiModalEntity to allowed kind of entities


benlabbe (approved):
OK

PR #554, opened 2025-05-27
allow SAM2 tracking

BertrandRenault:
## Description

Allow tracking with Pixano Inference.

TODO:
- [x] choose number of frames to track: setting in model config panel.
- [x] better handling of image vs video tracking in SelectModel
- [x] better handling for launching request (option to validate before segmentation, which allows to put several +/- points)
- [x] tests back


benlabbe (approved):
I am not able to test this PR. But I saw it working in several demos.

PR #573, opened 2025-07-02
feat(front): Improve navigation keyboard shortcuts

cpvannier:
## Description

* Since a lot of annotation actions in Pixano involve using the mouse in the right hand, add the option to navigate video frames with the left hand (WASD/ZQSD) rather than just with the right hand (Arrow keys)
* Replace `event.code` instead of `event.key` to be able to use the same physical keys regardless of keyboard layout (QWERTY, AZERTY, or other)

### DIFF
--- a/ui/apps/pixano/src/components/layout/DatasetItemHeader.svelte
+++ b/ui/apps/pixano/src/components/layout/DatasetItemHeader.svelte
@@ -42,7 +42,7 @@ License: CECILL-C
           break;
         case "ArrowLeft":
         case "KeyA":
-      await goToNeighborItem("previous");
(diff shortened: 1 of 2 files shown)


BertrandRenault:
Nice one !
I would suggest to use same keys also for DatasetItems navigation (shift+arowLeft/Right) (see https://github.com/pixano/pixano/blob/68b4a6b89dfb5814f071373731b92eb0978d2080/ui/apps/pixano/src/components/layout/DatasetItemHeader.svelte#L37), and maybe adapt Tooltip for this one (same file, L.60 & 66)

BertrandRenault (approved):

PR #577, opened 2025-07-07
add trailing slash where redirect occurs

BertrandRenault:
## Issue

sometime fetch cause a redirection because of a missing trailing '/' in url

## Description

Add trailing slash for indentified redirections


benlabbe (approved):
Hope this will fix the deployment behind an HTTPS reverse-proxy

PR #578, opened 2025-07-07
temporary textSpan on creation

BertrandRenault:
## Description

Create a temporary TextSpan while creating a TextSpan (button "TagSelectedText") for better ergonomy.
It is removed on cancel (with onDestroy) or submit.


benlabbe (approved):
Nice improvement

PR #580, opened 2025-07-08
add unsubscribers

BertrandRenault:
## Issue

After using Pixano for some time (navigating through several datasetItems and datasets), we can see in logs that some things that should be unique are run several times. 

## Description

A potential origin of this behaviour could be because pixano front doesn't usually unsubscribe, though there is a consequent number of subscribe in front code.

By adding all unsubscribers (called in "onDestroy"), we hope ths could resolve the issue.

WIP ...still under testing...

Note: 
In SceneInspector - line 48 :  $: mediaViews.subscribe(...);
This one cause a bug if unsubscribed... need further investigation...

### DIFF
--- a/ui/apps/pixano/src/components/dataset/DatasetPagination.svelte
+++ b/ui/apps/pixano/src/components/dataset/DatasetPagination.svelte
@@ -6,6 +6,8 @@ License: CECILL-C
 
 <script lang="ts">
   // Imports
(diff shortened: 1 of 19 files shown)


benlabbe (dismissed):
Sorry . I cannot test this unsubscribers. I don't have the protocol to test the evolution.

PR #582, opened 2025-07-10
TextSpan section as table + abort on fetch items_info

BertrandRenault:
## Issue

1. TextSpan section on ObjectCard is not so clear, and we could need more details than "role" (need to add "concept")
2. items_info (for DatasetLibrary preview cards tooltip) could delay model loading if dataset clicked before all fetches done.

## Description

1. Use an HTML table for display, for more clarity and display of column names
2. use fetch AbortSignal to cancel ongoing fetches when leaving Dataset Library page.


benlabbe (approved):
Great improvement

PR #583, opened 2025-07-10
add timeout to pixano inference connect

BertrandRenault:
## Issue

If for some reason internet connection is broken, inference/connect fetch would hang for default timeout (depending on browser, could take some time)

## Description

Set a shorter timeout (4sec) (this is a fast request on pixano-inference server side)


benlabbe (approved):
ok

PR #585, opened 2025-07-16
Table counts

BertrandRenault:
## Description

Add counts of objects and shapes (from items_info) in dataset explorer
image


BertrandRenault:
> Can you confirm that the names of the columns are derived by the names of the tables in the database schema ? So in dataset with classical entities (not tracks) , we would read 'objects' or 'multimodal_entites' . Is it ?

Yes, the column names are took from "annotations" and "entities" subfield of returned items_info route. So yes, this are the table names in the current dataset.

benlabbe (approved):
Great new feature. Congratulations

PR #586, opened 2025-07-16
animated save button + equality fixes

BertrandRenault:
## Description

Save button now show an animated pulsing dot when there is something to save.

+ fixes some == vs === when relevant.


benlabbe (approved):
Great

Issue #491, opened 2025-03-04
Rework FeaturesValues


BertrandRenault:
## Feature description

"FeaturesValues" feature (set of predefined values for features) is outdated, and may have issue if some custom features name are not unique.

## Expected solution

- This need a rework to be more robust. (allow non unique features name from different tables)
- features_values.json should be more concise and user friendly.
- the ability to extend with input values (for non restricted) is broken.
- lookup of existing values is broken (has been removed). Put it back ?

PR #589, opened 2025-07-17
allow Literal, extract FV from them, add_constraint

BertrandRenault:
## Description

- [x] allows to use Literal in DatasetItem fields
- [ ] allows Literal for DaseSchema fields too, not only DatasetItem
- [x] extract features values from Literal -- DatasetItem fields
- [ ] extract features values from Literal -- non DatasetItem fields
- [x] add_constraint method to add/update constraints

Do not resolve, but some step toward #491

### DIFF
--- a/pixano/datasets/builders/dataset_builder.py
+++ b/pixano/datasets/builders/dataset_builder.py
@@ -15,7 +15,7 @@ import shortuuid
 import tqdm
 from lancedb.table import Table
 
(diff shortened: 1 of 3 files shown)


BertrandRenault:
LanceDB convertion to Arrow type doesn't support Literal.
Managing this is too complex for intended usage.
We will simply use a Dataset method too add/update constraints (FeaturesValues)
Closed

PR #590, opened 2025-07-17
add_constraint method

BertrandRenault:
## Description

Add a method in Dataset class: add_constraint to allows to set allowed values (DatasetFeaturesValues)

note: (scope: front) is just a small fix linked to an incorrect field name in DatasetFeaturesValues.


benlabbe (approved):
Nice feature

PR #593, opened 2025-07-22
fix selectmodelmodal

BertrandRenault:
## Issue

since #571 the Select Model modal opens too frequently (after a model has been loaded once, each dataset item change, or tool change, will open it)

## Description

fix with more complete tests


benlabbe (approved):
OK

PR #656, opened 2026-06-03
fix: TextSpan missing view_id on save

benlabbe:
fix: TextSpan missing view_id on save, EntityCard reactive stalls, BBox resize aspect ratio lock, and ChildCard typography consistency

- entityOperations.ts: add view_id to TextSpan data payload so freshly drawn spans are grouped and rendered correctly after save
- EntityCard.svelte: add void entities.value reactive bridge to allowedChilds, totalAllowableChildCount, and annotationTypeSummary deriveds so annotation deletion immediately removes the card from DOM
- BBox2D.svelte: add keepRatio={false} to Konva Transformer to allow independent width/height adjustment during corner resize
- ChildCard.svelte: restructure bbox coordinate grid into two flex columns with a 50px vertical divider for clearer visual separation
- ChildCard.svelte: normalize key-value typography across all annotation

### DIFF
--- a/ui/apps/pixano/src/components/workspace/Inspector/ChildCard.svelte
+++ b/ui/apps/pixano/src/components/workspace/Inspector/ChildCard.svelte
@@ -440,17 +440,32 @@ License: CECILL-C
         {#if bboxData}
           <div class="rounded-md bg-muted/30 border border-border/30 p-2 space-y-1.5">
             <!-- Coordinate grid -->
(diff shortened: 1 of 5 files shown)


benlabbe:
@jrabary  @vmulet-benzo  @glapouge  ? could you approve this PR  ?

vmulet-benzo (approved):

PR #682, opened 2026-07-06
feat: add the COCO importer and exporter with round-trip

jrabary:
## Issue

P3.1 of the 0.8.0 redesign (spec §7.2). Based directly on `arch/data-import` — no stack.

## Description

The second real format lands on the io core:

- **`CocoImporter`** — two-pass streaming: pass 1 streams `annotations` into an index keyed by `image_id` (**SQLite spill** above a row threshold, so annotation files larger than RAM import in bounded memory); pass 2 streams `images` joining the index. `ijson` (the new `coco` extra) enables true streaming; without it the file loads whole with a warning Finding at analyze — never silently.
- **Intrinsic schema** via a new `DatasetImporter.resolve_info` hook: COCO knows its own schema (entity `category`/`supercategory`/`iscrowd` + bbox/mask/keypoint tables), while a user-declared schema block still wins. LeRobot will use the same hook.
- **Mappings**: absolute `xywh` boxes normalize against image dims; polygon segmentation → `CompressedRLE.from_polygons`, uncompressed RLE dicts → `from_urle`, compressed counts pass through; keypoints → the `coco-17` template with `v={0,1,2}` → invisible/hidden/visible; `file_name` resolves under `<split>/`, `image/<split>/`, `images/<split>/`; **uri mode** stores `coco_url` verbatim (missing url is an analyze error, per the §6 media contract).
- Deterministic ids, resume cursor `{split, image_ordinal}`, detection on `instances_*.json`, analyze validates media existence and category references. Grid previews come free from the engine (#680).

8 tests: fixture end-to-end (exact counts, idempotent rerun, namespaced ids), **spill path id-identical to the in-memory path**, cursor resume, detection, synthetic keypoints+polygons, unknown-category analyze error, uri mode. Full suite green (585).

Next: the COCO exporter + round-trip (P3.2), then LeRobot with the frames-to-SequenceFrame default.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -49,6 +49,9 @@ dependencies = [
   "pyyaml>=6.0,<7.0",
 ]
 
+[project.optional-dependencies]
+coco = ["ijson >= 3.2, < 4.0"]
+
 [dependency-groups]
 docs = [
   "griffe >= 1.0.0",
--- a/src/pixano/datasets/io/__init__.py
+++ b/src/pixano/datasets/io/__init__.py
@@ -93,7 +93,9 @@ __all__ = [
 
 
 # Built-in formats register on package import (entry-point plugins load lazily).
+from .formats.coco.importer import COCO  # noqa: E402
 from .formats.pixano_jsonl.importer import PIXANO_JSONL  # noqa: E402
 
 
 FORMATS.register(PIXANO_JSONL)
+FORMATS.register(COCO)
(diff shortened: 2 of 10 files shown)


jrabary:
Second commit completes the PR as planned (P3.1 + P3.2): the **COCO exporter** — reader-driven single pass emitting `instances_<split>.json` + media under `image/<split>/` (the importer's own lookup layout, so exports re-import with zero flags). Categories are now actually populated (deterministic alphabetical ids — v1 exported them empty), boxes denormalize, masks emit compressed RLE, keypoints map back to COCO visibility triplets, and `pixano_id` extensions keep every image/annotation traceable to its source row. `export_dataset` dispatches both `pixano_jsonl` and `coco`.

Round-trip test: fixture import → COCO export → re-import preserves every table count, the category set, and bbox coords to 1e-3. Full suite green (586).

PR #683, opened 2026-07-06
feat: add the LeRobot importer with frames-to-SequenceFrame default

jrabary:
## Issue

P3.4 of the 0.8.0 redesign (spec §7.3), with the agreed re-scope: **frame extraction is the default** so episodes are annotatable in the 0.8 UI — the Video-window path stays as an opt-in for browse-scale datasets. Based on `arch/data-import`; independent of #682.

## Description

- **Layout parsing, pyarrow-only** (no `lerobot` dependency): v2.1 (`meta/episodes.jsonl`, one video per episode) and v3 (episodes metadata parquet with per-camera `from/to_timestamp` windows into shared shards), both via the `video_path` templates in `meta/info.json`.
- **Default `frames: "extract"`** — ffmpeg decodes each episode's window at the dataset fps into embedded `SequenceFrame` rows (the DAVIS pattern: browse grid, timeline, bbox/mask/tracklet annotation all work today; previews stamp automatically). Scale guards: `max_frames_per_episode` uniform-stride cap and a JPEG **size estimate in the analyze plan** before anything is written.
- **Opt-in `frames: "reference"`** — `Video` time-window rows (v3 windows; v2.1 degenerate whole-file windows) for large datasets browsed without annotation.
- **`episodes` selection** (list or `"start:end"`), records carry `episode_index`/`tasks`/`length`, resume cursor per episode, deterministic ids.
- Analyze findings: `ffmpeg_required`, **AV1 codec warning** (extracts fine, Safari playback limitation documented), missing shards.
- `DatasetImporter.resolve_info` now receives the source, so source-dependent schemas (one view per camera) resolve cleanly — same hook COCO uses.

Options ride in `dataset.yaml`/spec `options:`, e.g. `options: {episodes: "0:9", max_frames_per_episode: 300}`.

10 tests over synthetic v2.1/v3 fixtures built from the sample video — including the window-math check that two episodes sharing one v3 shard produce disjoint frame timestamp ranges. Full suite green (596).

Remaining LeRobot scope (follow-up): `hub://` meta-only mode via `huggingface_hub`.

### DIFF
--- a/src/pixano/datasets/io/__init__.py
+++ b/src/pixano/datasets/io/__init__.py
@@ -27,7 +27,15 @@ from .errors import (
 from .ids import IdLedger, namespace_prefix, stable_id
 from .importer import BatchBundle, Cursor, DatasetImporter, DetectResult, SourceRef
 from .manifest import ImportManifest
-from .media import MediaResolver, ResolvedMedia, VideoProbe, ffprobe_available, probe_image, probe_video
+from .media import (
+    MediaResolver,
+    ResolvedMedia,
+    VideoProbe,
+    ffmpeg_available,
+    ffprobe_available,
+    probe_image,
+    probe_video,
+)
 from .plan import AnalyzeLimits, Finding, ImportPlan, PreflightReport, Provenance, SamplePreview
 from .progress import ProgressEvent, ProgressSink, ThrottledSink, TqdmSink
 from .reader import RecordBundle, RecordBundleReader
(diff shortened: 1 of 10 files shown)


jrabary:
Added P3.6 as a second commit: slow-marked scale/memory tests (import RSS delta bounded on a 160MB payload, 10x-rows-under-30x-time linearity, export blob streaming under an 8MB reader budget, reader query count scales with pages not records). CI deselects them; locally they run in ~7s via `pytest tests/datasets/io/test_scale.py -m slow`.

PR #697, opened 2026-07-08
feat: import raw media folders and surface the resolved schema in plans

jrabary:
## Description

Backend half of the import-wizard redesign (PR-A of two). Two gaps close:

**Raw media import** — `pixano_jsonl` media-only mode loses its single-image-view constraint and becomes the raw-data path the redesigned wizard needs (new `media_only.py` module):
- **Images in view folders**: `source/<view>/*.jpg` stem-match across views into multi-view records (missing stems are warnings); flat folders and `train/val/test` split folders both work; view names match folders exactly or snake_cased.
- **Videos**: extract frames to annotatable `SequenceFrame` rows by default (mirrors the LeRobot 0.8 decision), capped by `options.max_frames_per_video` with uniform stride; `options: {frames: reference}` opts into browse-scale `Video` rows. Analyze estimates the extracted JPEG size and flags `ffmpeg_required`.
- **Text files**: `.txt`/`.md` import as `Text` rows with embedded content; `classification`/`text_span` slots seed by default.
- Discovery is one deterministic function shared by analyze, `resolve_info`, and ingest — the schema shown at plan time is exactly what ingest builds. When the spec declares no views, analyze infers them from the folder shape; ambiguous layouts (mixed media kinds, split names mixed with view folders, root files next to view dirs) are findings, never guesses. Ids stay identical to 0.8.0 for the existing flat single-view layout (regression-tested).
- Auto-detect now recognizes bare media folders at low confidence (0.3), so the wizard's "let Pixano detect it" works on raw folders.
- Metadata-less splits inside a real JSONL source keep the original single-image-view rule (unchanged contract).

**`plan.inferred_schema` populated** — `analyze()` in `io/api.py` now stamps the resolved schema on every plan via a new `serialize_dataset_info_schema()` helper (same `{base, name?, fields}` manifest dialect `GET /datasets/{id}/info` already serves). This is the spec §4 behavior ("analyze infers … and embeds the inferred spec in the ImportPlan for confirmation") that was declared but never wired. Covers all formats — pixano_jsonl, COCO (implied category attrs), LeRobot (camera views) — over REST, Python, and CLI. `pixano data import --dry-run` renders the schema block:

```
Schema (workspace: image):
- View 'left': Image
- View 'right': Image
- Entity attrs: category: str
- Annotations: bbox, keypoint
```

Spec doc updated (§5 media-only section). Frontend consumption (intent-first wizard, schema builder, schema panel) follows in PR-B.

## Tests

- New `tests/datasets/io/formats/test_media_only.py` (19 tests): discovery rules and every ambiguity finding, multi-view stem matching, split nesting, video extract (capped) + reference modes, text content, id-stability of the legacy layout, idempotent add-mode reruns, inferred-schema content, raw-folder detection.
- `tests/app/test_api_io.py`: `inferred_schema` asserted on the REST contract + a full raw-folder analyze→import flow with UI-built schema (custom entity attr, chosen annotation slots).
- Full suite: 676 passed / 11 skipped; io subset 233 passed. Verified live: CLI dry-run schema rendering, extract/reference ingest, add-mode idempotency.


jrabary:
Added `GET /io/browse` — lists a directory's subfolders (dirs only, hidden skipped, capped at 500) with shallow source hints (dataset.yaml → pixano, meta/info.json → lerobot). Backs the wizard's new Browse… source picker in #698. Since Pixano runs on the user's machine, this browses their local filesystem; it exposes nothing /io/analyze doesn't already accept. API tests included.

jrabary:
**Correction — deployment model.** In production, users run the UI in a browser on their own machine and have no access to the server filesystem, so the earlier `GET /io/browse` was the wrong model and is now **removed**. Replaced by **staged client uploads**:

- `POST /io/uploads` creates a staging session under the engine state dir (`<data_dir>/.pixano/uploads/<id>/`)
- `PUT /io/uploads/{id}/files/{rel_path}` streams raw file bodies preserving the folder structure (path traversal rejected; no python-multipart dependency needed)
- `DELETE /io/uploads/{id}` discards a session
- Imports run against the staged folder unchanged (raw discovery, schema inference, resume). The session is removed eagerly once its job is done or cancelled (errored/interrupted keep it for resume); `boot_recover` reclaims stale orphans after 24h.

API tests cover the full upload→analyze→import→GC flow, traversal rejection, and boot GC. 18/18 locally; io regression 233/233.

PR #717, opened 2026-07-26
fix: working column settings, 'More like this' affordance, embeddings health

jrabary:
## Issue

Fixes #

## Description

Three defects from live testing of the explorer redesign. **Stacked on #716** (retarget along the chain as PRs below merge).

### 1. Column settings now actually work (pre-existing bug)

The settings popover looked functional but changed nothing: `createSvelteTable` eagerly spread `...options.state`, freezing the Svelte `$state` getters into plain values — TanStack's render-time `getState()` never touched a signal, so the template had no dependency on column order/visibility and never re-rendered. The adapter now merges state through a **lazy getter-forwarding proxy** (the canonical svelte5-tanstack pattern), so hide/reorder update **in place**. The abandoned `svelte-sortable-list` dependency is gone (its "reorder" actually *swapped* items, and native drag is unreliable inside a Popover portal) — replaced by keyboard-accessible **move up/down buttons**.

### 2. "More like this" replaces the Sparkle AI icon

Vector search isn't an AI action; the sparkle misled. Adopted the battle-tested "similar images" vocabulary (Pinterest / Google Images): the **`Images` glyph + "More like this"** label — tooltip on the table row action, **labeled pill buttons** ("Open" / "More like this") on the card hover overlay, `Images` on the "Similar to" chip. Sparkle is removed from every search affordance; the search box keeps the magnifier (the labeled **Text | Semantic** toggle communicates the mode).

### 3. Embeddings health + repair (fixes the advertised-but-500 deadlock)

The sidecar file alone both advertised semantic search **and** blocked recompute (the job read the phantom table first) — once the table broke, the dataset was stuck: advertised, 500 on search, recompute erroring.

- **`Dataset.record_embedding_health()`** — cheap metadata check → `ready | partial | empty | missing_table | dim_mismatch | corrupt`; **`drop_record_embeddings()`** for clean removal.
- **Load hardening**: a corrupt sidecar (e.g. `dim<=0`) previously raised out of `Dataset.__init__` and 500'd **every endpoint** of the dataset; it now degrades to "no embeddings".
- **`/filters`** advertises semantic only when healthy and ships `status`/`detail`/coverage; **search** short-circuits degraded stores with actionable **503/409 — never 500** (including a query-vs-stored dim check for model drift).
- **`force` recompute** drops and rebuilds the store (the repair path); without force, a model switch or a resume over a broken store is refused with a 409 (stops silent embedding-space mixing).
- **UI**: a warning-styled **"Repair semantic search"** button replaces the toggle when the store is degraded; **"Update embeddings"** appears when coverage is partial; the failure banner now shows the backend's actionable detail instead of always blaming the inference server.

### Tests
Backend **858 passed** — 17 new health/drop unit tests + 7 new corruption API tests (missing table → 503 + repair status; dim mismatch → 409; force recompute end-to-end repair; model-switch guard). Frontend **214** · check 0 errors · lint · format · build.

### Live smoke
Delete `db/embeddings.lance` (keep `embeddings.json`) on an embedded dataset → toolbar shows **Repair semantic search** (no 500 anywhere) → click → job rebuilds → semantic works again. Toggle/reorder columns in the popover → table updates instantly. Hover a card → labeled "Open" / "More like this" buttons.


jrabary:
Heads-up on scope after the retarget: **#716's content never actually reached `releases/0.8`** — #715 was squash-merged to `releases/0.8` first, and #716 was then merged into its (now-orphaned) base branch `feat/explorer-redesign`, so the gallery never flowed to the release branch.

This PR now carries those commits too: after the rebase it contains the **gallery view + resizable previews + text excerpts (#716)** plus the fixes and the semantic-only search bar. Merging this single PR into `releases/0.8` delivers everything.

